PR #9, opened 2018-10-07
Continue running if github commit hash not valid

20lives:
If provided invalid GitHub commit hash the test session will not start.
this commit will allow it to continue running without github status checks

Signed-off-by: Lior Haim <lior@daynix.com>

### DIFF
--- a/lib/github.rb
+++ b/lib/github.rb
@@ -3,7 +3,7 @@ require 'octokit'
 # github class
 class Github
   def initialize(config, project, commit)
-    @up = false
+    @api_connected = false
     @logger = project.logger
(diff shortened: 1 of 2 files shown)


lib/github.rb
```diff
@@ -32,6 +32,7 @@ def find_pr
     pr = @github.pulls(@repo).find { |x| x['head']['sha'] == @commit }
     if pr.nil?
       @logger.error('Pull request commit hash not valid')
+      @up = false
```
YanVugenfirer: Please change the naming to something that is semantically correct.

20lives:
Updated.

PR #27, opened 2018-11-27
Abstraction for uploading the results

bish22ah:
Added a result_uploader field in the config.json:
    "studio_username": "Administrator",
    "studio_password": "Qum5net.",
 -> "result_uploaders": [ "dropbox" ]
}
Array of the wanted result uploaders to function during a run.

Changes:

AutoHCK's Project instance has a ResultUploader instance instead of a
Dropbox instance and the ResultUploader uses the ruby factory pattern to
implement the result uploaders abstraction.

At the moment, dropbox is configured to be the sole result uploader,
but this is the first step towards supporting more result uploaders.


bish22ah:
Closed, sending new PR asap.

PR #29, opened 2018-11-29
Fix TERM signal handling and graceful shutdown

Basils:
This patch fixes issues where TERM signal was not handled properly and it now checks for leftovers.

### DIFF
--- a/lib/virthck.rb
+++ b/lib/virthck.rb
@@ -124,4 +124,9 @@ class VirtHCK
     s_id = @id.to_s.rjust(4, '0')
     `ps -A -o cmd | grep '[\-]name HCK-Client#{id}_#{s_id}'`.split("\n").any?
   end
+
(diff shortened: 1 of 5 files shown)


lib/virthck.rb
```diff
@@ -124,4 +124,9 @@ def client_alive?(name)
     s_id = @id.to_s.rjust(4, '0')
     `ps -A -o cmd | grep '[\-]name HCK-Client#{id}_#{s_id}'`.split("\n").any?
   end
+
+  def studio_alive?
+    s_id = @id.to_s.rjust(4, '0')
+    `ps -A -o cmd | grep '[\-]name HCK-Studio_#{s_id}'`.split("\n").any?
```
20lives: Maybe we can create a base method for checking live machines and extend it for client_alive? and studio_alive?

PR #53, opened 2019-05-22
Id allocation retry in case generator is busy

Basils:
Signed-off-by: Basil Salman <Basil@daynix.com>

### DIFF
--- a/lib/virthck.rb
+++ b/lib/virthck.rb
@@ -16,9 +16,10 @@ class VirtHCK
 
   def assign_id
     @id = @id_gen.allocate
-    if @id.negative?
-      @logger.fatal('No available ID, wait for a test session to end')
-      exit 1
+    while @id.negative?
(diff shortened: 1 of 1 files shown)


lib/virthck.rb
```diff
@@ -15,11 +15,7 @@ def initialize(project)
   end
 
   def assign_id
-    @id = @id_gen.allocate
-    if @id.negative?
-      @logger.fatal('No available ID, wait for a test session to end')
-      exit 1
-    end
+    @id = @id_gen.allocate while @id.negative?
```
YanVugenfirer: I think we need to wait for a while here. Otherwise, we just are in the loop and blowing up the log. Let's say next retry in 20 seconds (use constant or create config value)

Basils: Fixed it with proper logging too

PR #55, opened 2019-06-03
Dev bish22ah

bish22ah:
1. Prevent NotReady state to Clients
2. Dropbox class updates
3. Custom exception AutoHCKError class
4. Better logging across multiple classes
5. Added exceptions to handle failed VirtHCK class runs
6. Updating return_when_client_up mechanism to use Thread's built in timeout
7. Studio class added StudioConnectionError exception class
8. Retry if VirtHCK orchestration failed
9. Disable TERM signal handling in case of a fatal error

### DIFF
--- a/bin/auto_hck
+++ b/bin/auto_hck
@@ -53,6 +53,7 @@ Signal.trap('TERM') do
 end
 
 Thread.abort_on_exception = true
+Thread.report_on_exception = false
 
 begin
   @project = Project.new(options)
--- a/bin/auto_hck
+++ b/bin/auto_hck
@@ -105,6 +105,9 @@ begin
   @client1.run_tests
   @client1.create_package
 rescue StandardError => e
+  Signal.trap('TERM') do
+    @project.logger.warn('SIGTERM(*) received, ignoring...')
+  end
   @project&.log_exception(e, 'fatal')
   clean_threads
   raise e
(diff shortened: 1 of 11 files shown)


lib/client.rb
```diff
@@ -208,6 +214,13 @@ def alive?
     false
   end
 
+  def prevent_not_ready
```
YanVugenfirer: I would rename to "reset_to_ready_state" or something similar.

lib/tests.rb
```diff
@@ -115,6 +115,11 @@ def handle_finished_tests(tests)
     print_tests_stats
   end
 
+  def prevent_clients_not_ready
```
YanVugenfirer: Same here: reset_clients_to_ready_state or clients_reset_to_ready_state or if this is client class already just reset_to_ready_state

bin/auto_hck
```diff
@@ -53,6 +53,7 @@ Signal.trap('TERM') do
 end
 
 Thread.abort_on_exception = true
+Thread.report_on_exception = false
```
YanVugenfirer: "report_exception?" probably better

bish22ah: This only prevents the thread exception to be reported to stderr, the exception will still be raised from the thread to the main thread.

lib/client.rb
bish22ah: On it...

lib/tests.rb
bish22ah: On it...

PR #63, opened 2019-09-05
Added "DF - Embedded Signature Verification Test (Certification)" to …

Basils:
…blacklist

This test runs only on HCK, it fails before running and breaks clients
readiness.

Signed-off-by: Basil Salman <basil@daynix.com>


20lives:
Maybe someone should implement what I have started working on: **Blacklists by platform #42**.
this would be a much better solution instead of just blacklisting this test on all systems.

PR #70, opened 2020-04-06
AutoHCK and Engine configuration separation

Basils:
Signed-off-by: Basil Salman <Basil@daynix.com>
Signed-off-by: Basil Salman <bsalman@redhat.com>


lib/project.rb
```diff
@@ -108,8 +108,6 @@ def validate_paths
 
   def normalize_paths
     @driver_path.chomp!('/')
-    @config['images_path'].chomp!('/')
```
YanVugenfirer, line 111: Don't we need something like this in virthck.rb code?

engines/virthck.rb
```diff
@@ -1,19 +1,28 @@
...
 require './lib/exceptions'
 
 # Virthck class
 class VirtHCK
+  VIRTHCK_CONFIG_JSON = './engines/virthck.json'
   def initialize(project, id)
     @project = project
     @logger = project.logger
-    @config = project.config
+    @config = read_json(VIRTHCK_CONFIG_JSON)
```
YanVugenfirer, line 12: Let's add a task to add a config validation.

lib/project.rb
Basils: We have something like this under /engines/virthck.rb

PR #81, opened 2020-11-02
Some fixes

Basils:


lib/setupmanagers/virthck/virthck.rb
```diff
@@ -104,6 +104,7 @@ def temp_file
   end
 
   def run(name, first_time = false)
+    sleep(rand(10))
```
YanVugenfirer, line 107: I would add why this is needed.

YanVugenfirer (approved):

PR #82, opened 2021-02-03
[RFC] AutoHCK: Auto image creation

kostyanf14:
This PR just for comments, many commits is not fully finished.


lib/project.rb
```diff
@@ -58,7 +61,12 @@ def init_multilog(debug)
   end
 
   def append_multilog
-    @logfile_path = "#{workspace_path}/#{tag}.log"
+    if @install.nil?
+      @logfile_path = "#{workspace_path}/#{tag}.log"
```
YanVugenfirer, line 65: Get paths from engine.

lib/project.rb
```diff
@@ -82,8 +96,14 @@ def init_class_variables(options)
     @timestamp = create_timestamp
     @tag = options.tag
     @driver_path = options.path
-    @driver = find_driver
-    @engine = @config['engine']
+    if options.install
```
YanVugenfirer, line 99: The usage of engine or engine specific options here should follow object-oriented principles.
The engine object or class should provide all the needed info.

Explicit driver installation here is actually the only place where this paradigm is broken now. 
Easy to fix to renaming the operation to more generic and again taking it from the engine.

YanVugenfirer (changes requested):

lib/engines/hcktest/hcktest.rb
```diff
@@ -6,7 +6,7 @@
 
 # HCKTest class
 class HCKTest
-  PLATFORMS_JSON = 'lib/engines/hcktest/platforms.json'
```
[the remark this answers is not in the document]
kostyanf14: platforms is also needed for install engine, so for now this is not fully specific.

lib/project.rb
kostyanf14: I think we should create init_class_variables function in engine and call it. What do you think?

lib/project.rb
kostyanf14: Will change

YanVugenfirer:
The question should then "install" be a separate engine? And by the way, if
it is is then it is "qemu_vm_install" engine.

YanVugenfirer:
On Wed, Feb 3, 2021 at 5:23 PM Kostiantyn Kostiuk <notifications@github.com>
wrote:

> *@kostyanf14* commented on this pull request.
> ------------------------------
>
> In lib/project.rb
> <https://github.com/HCK-CI/AutoHCK/pull/82#discussion_r569503417>:
>
> > @@ -82,8 +96,14 @@ def init_class_variables(options)
>      @timestamp = create_timestamp
>      @tag = options.tag
>      @driver_path = options.path
> -    @driver = find_driver
> -    @engine = @config['engine']
> +    if options.install
>
> I think we should create init_class_variables function in engine and call
> it. What do you think?
>
Good idea.

kostyanf14:
>The question should then "install" be a separate engine? And by the way, if
>it is is then it is "qemu_vm_install" engine.

Install engine has many differences from the hcktest:
- virthck command line
- second iso preparation
- does not need driver path

I think better to split it.

about naming this is not qemu_vm_install, because only virthck setup manager know about QEMU, install engine will call setup manger for command line generation. If we implement new setup manager for Hyper-V, we still can use this install engine.

YanVugenfirer:
On Wed, Feb 3, 2021 at 5:33 PM Kostiantyn Kostiuk <notifications@github.com>
wrote:

> The question should then "install" be a separate engine? And by the way, if
> it is is then it is "qemu_vm_install" engine.
>
> Install engine has many differences from the hcktest:
>
>    - virthck command line
>    - second iso preparation
>    - does not need driver path
>
> I think better to split it.
>
OK

> about naming this is not qemu_vm_install, because only virthck setup
> manager know about QEMU, install engine will call setup manger for command
> line generation. If we implement new setup manager for Hyper-V, we still
> can use this install engine.
>
But we are installing HCK specific stuff, right?
What if we have iperf engine that runs with virthck setup manager? Will we
have a different setup manager in such case as well?

kostyanf14:
>But we are installing HCK specific stuff, right?
>What if we have iperf engine that runs with virthck setup manager? Will we
>have a different setup manager in such case as well?

From my point of view there are two variant:
1. Yes, this is HCK specific. So we should create hckinstall engine. For iperf we will create iperf install engine, iperf test engine and we can use virthck setup manager if run iperf over QEMU.
2. we create some scripts for iperf VM preparation, and install engine call these scripts instead of HCK setup scripts. We only need iperf test engine.

kostyanf14:
This PR is no longer needed. New PR will be created.

PR #84, opened 2021-02-17
Move all code into AutoHCK module (namespace)

kostyanf14:
Please look in commit messages for more details.

Tested on Groan server

@Basils Should we test these changes in other servers?

### DIFF
--- a/lib/version.rb
+++ b/lib/version.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
-class AutoHCK
-  VERSION = '0.6.2'
+# AutoHCK module
+module AutoHCK
+  class AutoHCK
+    VERSION = '0.6.2'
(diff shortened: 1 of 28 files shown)


Basils (approved):

lib/version.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
-class AutoHCK
-  VERSION = '0.6.2'
+# AutoHCK module
+module AutoHCK
+  class AutoHCK
+    VERSION = '0.6.2'
```
YanVugenfirer, line 6: I think such changes should the propagate version.

kostyanf14: Added commit with the version bump

YanVugenfirer (approved):

PR #85, opened 2021-02-22
AutoHCK new features

kostyanf14:
Please see the commits message for more details

### DIFF
--- a/bin/auto_hck
+++ b/bin/auto_hck
@@ -4,7 +4,6 @@
 require 'filelock'
 require './lib/cli'
 require './lib/project'
-require './lib/engines/engine'
 
 # AutoHCK module
 module AutoHCK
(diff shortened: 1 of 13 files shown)


YanVugenfirer (approved):

bin/auto_hck
```diff
@@ -16,7 +17,8 @@ module AutoHCK
   AUTOHCK_RETRIES = 5
 
   at_exit do
-    @engine&.close
+    @project&.setup_manager&.close
+    @project&.engine&.close
```
Basils, line 21: if engine is initialized in the scope of project and setup_manager is initialized in the scope of engine, I think it is better to let project and engine handle the closing accordingly.

kostyanf14: Yes, you a right. Will fix this

YanVugenfirer: @Basils Any more comments?

PR #86, opened 2021-03-01
Initial implementation of image creation

kostyanf14:
The current implementation has a few limitations, that will be fixed in the next PR
- ~~install engine installs all images (studio and 2 clients), existing studio image will be used unless - **--force option is added**~~
- studio and client VMs will be installed from the one Windows ISO - **client platform will be parsed from filename**
- the corresponding version of HCK/HLK should be downloaded manually - **automatic download will be implemented**
- the corresponding windows ISO should be downloaded manually - **automatic download will be implemented**

### DIFF
--- a/lib/engines/hckinstall/hckinstall.rb
+++ b/lib/engines/hckinstall/hckinstall.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'uri'
+
+require './lib/auxiliary/json_helper'
+require './lib/auxiliary/host_helper'
+require './lib/auxiliary/iso_helper'
+require './lib/engines/hckinstall/setup_scripts_helper'
+
+# AutoHCK module
+module AutoHCK
+  # HCKInstall class
+  class HCKInstall
+    include Helper
+
+    attr_reader :platform
+
+    def initialize(project)
+      @project = project
+      @logger = project.logger
+      @project.append_multilog("#{project.install_platform}.log")
+      @logger.debug('HCKInstall: initialized')
+    end
+
+    def driver
+      nil
+    end
+
+    def run
+      @logger.debug('HCKInstall: run')
+    end
+
+    def close
+      @logger.debug('HCKInstall: close')
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


lib/engines/hckinstall/hckinstall.rb
```diff
@@ -15,23 +15,205 @@ class HCKInstall
 
     attr_reader :platform
 
+    PLATFORMS_JSON = 'lib/engines/hcktest/platforms.json'
+    ISO_JSON = 'lib/engines/hckinstall/iso.json'
+
+    ANSWER_FILES = ['autounattend.xml', 'unattend.xml'].freeze
+    CLIENTS = %w[c1 c2].freeze
+
+    STUDIO_INSTALL_TIMEOUT = 3 * 60 * 60 # 3 hour
```
YanVugenfirer, line 24: I think moving those parameters to JSON file is a better idea.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -15,23 +15,205 @@ class HCKInstall
...
+    CLIENT_INSTALL_TIMEOUT = 1 * 60 * 60 # 1 hour
```
YanVugenfirer, line 25: What's the difference?

YanVugenfirer (changes requested):

lib/engines/hckinstall/hckinstall.rb
kostyanf14: HLK studio required more time for installation. The longest studio installation time was 2 hours.
The longest client installation time was 30 min.

lib/engines/hckinstall/hckinstall.rb
kostyanf14: Will do this

kostyanf14:
Added hckinstall config file with all timeouts and unattended files list. Fixed first limitation (`--force-install` option added)

Basils (approved):

YanVugenfirer (approved):

PR #89, opened 2021-03-05
Load platform for the Studio VM from config file

kostyanf14:


lib/engines/hckinstall/studio_platform.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "kit": "HLK1809server",
```
YanVugenfirer, line 3: Why there are two kits for the same platform?

What about Windows 10 and previous OSes?

kostyanf14: >Why there are two kits for the same platform?

This file is used to get the Studio VM platform for the specific kit. We should install Server 2019 for HLK1809server and HLK1809 KITs

> What about Windows 10 and previous OSes?

Will update files with all other kits.

kostyanf14:
studio_platfroms.json updated

YanVugenfirer (approved):

PR #132, opened 2021-11-15
Update triggers logic

kostyanf14:


triggers.yml.example
```diff
@@ -1,40 +1,120 @@
...
+    - vioserial/lib/
+    - vioserial/libtestapp/
+    - vioserial/libtestsvc/
+    - vioserial/LICENSE
 viosock:
-  - 'viosock'
+  include:
+    - viosock/
+  exclude:
+    - viosock/ViosockPackage/
```
YanVugenfirer, line 94: Not sure about excluding the package.

triggers.yml.example
```diff
@@ -1,40 +1,120 @@
...
+  include:
+    - viostor/
+  exclude:
+    - viostor/LICENSE
+"*":
+  include:
+    - /
+    - Tools/
+    - VirtIO/
+    - packaging/
```
YanVugenfirer, line 110: Same here. I think packaging projects should be included.

YanVugenfirer (approved):

triggers.yml.example
YanVugenfirer: We discussed it off the list. Packaging projects should be added.

YanVugenfirer (changes requested):
Do not exclude packaging projects

triggers.yml.example
kostyanf14: Fixed

triggers.yml.example
kostyanf14: Fixed

YanVugenfirer (approved):

PR #134, opened 2022-01-05
Add CodeQL test as Github Action

YanVugenfirer:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '25 3 * * 2'
```
kostyanf14, line 21: Do we really need a scheduled job? I think push and PR is enough

YanVugenfirer: Yes, I also think that we don't need it. You are welcomed to changed, I just used the template  :)

kostyanf14 (approved):

PR #142, opened 2022-07-06
Add logs when test are queued or running very long time

kostyanf14:


lib/engines/hcktest/tests.rb
```diff
@@ -100,10 +105,8 @@ def check_test_duration_time
 
       return if diff < 2 * duration + time_to_seconds(RUNNING_TEST_TIMEOUT)
```
YanVugenfirer, line 106: Don't we want to take the time out from the test description?

YanVugenfirer (approved):

PR #143, opened 2022-07-25
Add VirtIO Viomem Driver support

kostyanf14:
@YanVugenfirer Please wait for merge untill get answer https://github.com/virtio-win/kvm-guest-drivers-windows/pull/794#issuecomment-1192287484

### DIFF
--- a/lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
+++ b/lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
@@ -1,0 +1,8 @@
+{
+  "name": "virtio-mem-pci",
+  "pluggable_memory_gb": 2,
+  "command_line": [
+    "-object memory-backend-ram,id=mem_backend,size=@pluggable_memory@",
+    "-device virtio-mem-pci@device_extra_param@,memdev=mem_backend,requested-size=1G,bus=@bus_name@.0"
+  ]
+}
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -224,6 +224,7 @@ module AutoHCK
       @drive_cache_options = []
       @define_variables = {}
       @run_opts = {}
+      @pluggable_memory_gb = 0
       @configured = false
     end
 
@@ -341,6 +342,16 @@ module AutoHCK
       }
     end
 
+    def memory_replacement_list
+      memory_gb = option_config('memory_gb')
+
+      {
+        '@memory@' => "#{memory_gb}G",
+        '@pluggable_memory@' => "#{@pluggable_memory_gb}G",
+        '@max_memory@' => "#{memory_gb + @pluggable_memory_gb}G"
+      }
+    end
+
     def options_replacement_list
       {
         '@machine_extra_param@' => @machine_extra_param.join,
@@ -358,7 +369,6 @@ module AutoHCK
         '@run_id_second@' => @id_second,
         '@client_id@' => @client_id,
         '@workspace@' => @workspace_path,
-        '@memory@' => "#{option_config('memory_gb')}G",
         '@cpu@' => option_config('cpu'),
         '@cpu_count@' => option_config('cpu_count'),
         '@cpu_model@' => option_config('cpu_model'),
@@ -367,6 +377,7 @@ module AutoHCK
         '@qemu_monitor_port@' => @monitor_port
       }.merge(config_replacement_list)
         .merge(machine_replacement_list)
+        .merge(memory_replacement_list)
         .merge(options_replacement_list)
         .merge(@define_variables)
     end
@@ -402,6 +413,8 @@ module AutoHCK
           var_value.merge! value
         when Array
           var_value << value
+        when Integer, String
+          instance_variable_set var, value
         else
           raise(QemuHCKError, "Variable #{var} has unsupported type")
         end
@@ -486,7 +499,7 @@ module AutoHCK
     def base_cmd
       [
         '@qemu_bin@ -enable-kvm -machine @machine_name@@machine_extra_param@ ',
-        '-m @memory@ -smp @cpu_count@,cores=@cpu_count@ ',
+        '-m @memory@,maxmem=@max_memory@ -smp @cpu_count@,cores=@cpu_count@ ',
         '-cpu @cpu_options@ -boot order=cd,menu=on ',
         '-nodefaults -no-user-config -usb -device usb-tablet -vnc :@vnc_id@ ',
         '-global kvm-pit.lost_tick_policy=discard -rtc base=localtime,clock=host,driftfix=slew ',
@@ -537,6 +550,7 @@ module AutoHCK
         '   VM VCPU .................... @cpu@',
         '   VM VCPUs ................... @cpu_count@',
         '   VM Memory .................. @memory@',
+        '   VM pluggable memory ........ @pluggable_memory@',
         '   VM display port ............ Vnc @vnc_id@/@vnc_port@',
         '   VM monitor port ............ Telnet @qemu_monitor_port@'
       ]
--- a/lib/engines/hcktest/drivers/viomem.json
+++ b/lib/engines/hcktest/drivers/viomem.json
@@ -1,0 +1,11 @@
+{
+  "name": "VirtIO Viomem Driver",
+  "device": "virtio-mem-pci",
+  "inf": "viomem.inf",
+  "install_method": "PNP",
+  "type": 0,
+  "support": false,
+  "reject_test_names": [
+    "Hardware-enforced Stack Protection Compatibility Test"
+  ]
+}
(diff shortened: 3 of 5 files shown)


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -193,6 +193,13 @@ def options_replacement_list
       }
     end
 
+    def double_memory
```
YanVugenfirer, line 196: Seems confusing name for function and variable. 
Maybe: memory_for_virtio_mem or pluggable_memory?

kostyanf14: No, this is just double memory. We run QEMU with option -m memory,maxmemory=double_memory. I hope that this reserve will be enough. While memory_for_virtio_mem is 1G in our configuration and pluggable_memory is the same as a run memory.

kostyanf14:
PR was rebased to current master

akihikodaki (commented):
Why do you need `double_memory`?

akihikodaki:
> Why do you need `double_memory`?

Nevermind. I read the help message of QEMU.

akihikodaki (changes requested):
That parsing is a bit tricky. Why don't you just convert `memory` into a normal JSON number so that we don't need parsing?

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "virtio-mem-pci",
+  "memory_extra_param": ",maxmem=@double_memory@",
```
akihikodaki, line 3: Why don't you always add `maxmem` and instead make it specific to virtio-mem-pci?

kostyanf14: In the case of NetKVM (3 VM), we will have the sum of maxmem greater than the host RAM on CI servers. I want to be sure that the host always has some free memory.

kostyanf14:
>That parsing is a bit tricky. Why don't you just convert memory into a normal JSON number so that we don't need parsing?

Do you suggest to convert 4G into 4294967296?

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
akihikodaki: Is it problematic?

akihikodaki:
> > That parsing is a bit tricky. Why don't you just convert memory into a normal JSON number so that we don't need parsing?
> 
> Do you suggest to convert 4G into 4294967296?

No, just 4.

kostyanf14:
> > > That parsing is a bit tricky. Why don't you just convert memory into a normal JSON number so that we don't need parsing?
> > 
> > 
> > Do you suggest to convert 4G into 4294967296?
> 
> No, just 4.

And suppose that we always use GB as a unit? Maybe rename key name to `memory_gb` in this case?

akihikodaki:
> > > > That parsing is a bit tricky. Why don't you just convert memory into a normal JSON number so that we don't need parsing?
> > > 
> > > 
> > > Do you suggest to convert 4G into 4294967296?
> > 
> > 
> > No, just 4.
> 
> And suppose that we always use GB as a unit? Maybe rename key name to `memory_gb` in this case?

Yes. Renaming key is a nice suggestion.

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
akihikodaki: I think it's not that strange that the sum of maxmem is greater than the host RAM.

Think of a multi-tenant machine farm; there are multiple machines, and each machine hosts multiple VMs. Also, assume that you want to scale VMs bigger without downtime.
In such a scenario, you set `maxmem` of all machines to match the maximum memory size a VM can hold. The sum of `maxmem` of VMs in a machine can surpass the physical memory size. Later, if too many memory is requested for a machine, some VMs on the machine will migrate to make space.
I think some cloud vendors already do this kind of trick.

kostyanf14: I don't know the QEMU source related to `maxmem` option. If you can confirm that QEMU will not allocate more than memory_size on Windows Guest without user action. I am ok.

akihikodaki: You can assume that; I believe the very purpose of `maxmem` is to support a scenario I have just described, and in that case, we should never have a problem with overcomitted `maxmem`.

kostyanf14: Ok. Let's assume that. We can always change this if see any problem.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -358,6 +358,7 @@ def full_replacement_list
         '@client_id@' => @client_id,
         '@workspace@' => @workspace_path,
         '@memory@' => "#{option_config('memory_gb')}G",
+        '@double_memory@' => "#{2 * option_config('memory_gb')}G",
```
akihikodaki, line 361: Let's have `@pluggable_memory@` and `@max_memory@`; `@max_memory` will be the sum of `@memory@` and `@pluggable_memory@`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: In this case, I need to talk with the driver developer. You can see that memory-backend-ram and virtio-mem-pci have a different memory size. What is `@pluggable_memory@` now `2G` or `1G`? That's why I don't prefer ideas with calculation.

akihikodaki: The same applies even now; changing the size of `mem_backend` into 512M will certainly cause a trouble. Having `@pluggable_memory@` only moves the place where the value `2G` is written, and allows calculating `@max_memory@`.

That said, I see no reason that `request-size` has to be inequal to `mem_backend` size.

kostyanf14: Just checked QE command line `extra_cmd= ' -object memory-backend-ram,id=vmem0,size=5G -device virtio-mem-pci,id=mem0,memdev=vmem0,node=0,requested-size=2G,indirect_desc=off,bus='+bus_num `

`request-size` != `mem_backend` size.

akihikodaki: It's better to ask if changing `requested-size` is OK.

kostyanf14:
@akihikodaki 
Let's continue with request-size != mem_backend size.

What comments do you have in this case?

akihikodaki:
> @akihikodaki Let's continue with request-size != mem_backend size.
> 
> What comments do you have in this case?

The comment in https://github.com/HCK-CI/AutoHCK/pull/143/files#r1495953291 still applies. I have nothing else to add.

akihikodaki:
> > > @akihikodaki Let's continue with request-size != mem_backend size.
> > > What comments do you have in this case?
> > 
> > 
> > The comment in https://github.com/HCK-CI/AutoHCK/pull/143/files#r1495953291 still applies. I have nothing else to add.
> 
> How do you expect to see the syntax of the @max_memory@ calculation? Or do you expect this as the internal logic of AutoHCK if @pluggable_memory@ is defined?

I expect an internal logic.

akihikodaki (changes requested):
Why don't you make `pluggable_memory_gb` a number defined with platform as `memory_gb` is?

kostyanf14:
>Why don't you make pluggable_memory_gb a number defined with platform as memory_gb is?

I am ok with making pluggable_memory_gb a number, but it doesn't relate to a platform. 
This fully depends on a device for now.

akihikodaki (changes requested):

akihikodaki:
> > Why don't you make pluggable_memory_gb a number defined with platform as memory_gb is?
> 
> I am ok with making pluggable_memory_gb a number, but it doesn't relate to a platform. This fully depends on a device for now.

I don't think `pluggable_memory_gb` is that different from `memory_gb`. The platform provides memory, and the devices and tests use it. Tests would just fail if the provided memory is not sufficient. We can use the same logic for pluggable memory as well. But It's not a strong opinion and I won't insist.

In any case, don't put it in `define_variables`. It is actually a list of string replacements and better to be typed as `T::Hash[String, String]`. Instead, put `pluggable_memory_gb` at the root of the device JSON.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -402,6 +413,9 @@ def load_device_info(device_info)
           var_value.merge! value
         when Array
           var_value << value
+        when Integer, String
+          @logger.info("Replacing #{var} value #{var_value} -> #{value}")
```
akihikodaki, line 417: Is this log really needed? We don't emit logs when overwriting values for `Hash`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: removed

akihikodaki (approved):

PR #144, opened 2022-08-02
Gemfile: Remove debase

akihikodaki:
debase depends on the internals of Ruby and does not work with Ruby 3.1.2. The debugger to use is a choice of each developer anyway so remove one from `Gemfile`.


Gemfile
```diff
@@ -20,7 +20,6 @@ gem 'sentry-ruby'
 gem 'sqlite3'
 
 group :development do
-  gem 'debase'
   gem 'ruby-debug-ide'
```
kostyanf14, line 23: Do we still need gem 'ruby-debug-ide' in this case? Maybe remove all development gems?

akihikodaki: I think they can be removed. Pushed commit ac08980a2eff01884a4fd396cab157f79e22fb94 to do so.

YanVugenfirer (approved):

kostyanf14 (approved):

PR #146, opened 2022-08-04
QemuHCK: Let studio manage network control bridge

akihikodaki:
Before this change, `QemuMachine` used to run the command to delete the network control bridge for each client. The later command executions was failing because the bridge was already removed with the earlier command and it did no longer exist.

With this change, the studio machine solely manages the network control bridge, and clients do no longer try to delete the network control bridge by its own. In fact, such a behavior was already present for the other network devices, and this change merely expands the scope of the behavior.

### DIFF
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -278,10 +278,11 @@ module AutoHCK
     def process_device(device_info)
       case device_info['type']
       when 'network'
-        dev, run_start, run_stop = @nm.test_device_command(device_info['name'], full_replacement_list)
+        dev = @nm.test_device_command(device_info['name'], full_replacement_list)
+        bridge_start, bridge_stop = @nm.test_bridge_command
         @device_commands << dev
-        @nm_commands_start << run_start
-        @nm_commands_stop << run_stop
+        @nm_commands_start << bridge_start
+        @nm_commands_stop << bridge_stop
       else
         cmd = device_info['command_line'].join(' ')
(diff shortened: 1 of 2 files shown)


kostyanf14 (approved):

YanVugenfirer:
@akihikodaki @kostyanf14 What happens during installation of the images? 
Awhile ago we had a mode where we code run only client. Is it still makes sense?

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -317,6 +314,10 @@ def process_hck_network
       @nm_commands_start << bridge_start
       @nm_commands_stop << bridge_stop
 
+      bridge_start, bridge_stop = @nm.test_bridge_command
```
kostyanf14, line 317: After these changes, we always will create a test bridge even not use it. Not sure if this is good or not.

kostyanf14:
@YanVugenfirer During the installation of the images we have 2 steps:
1) run only studio
2) run studio and clients

Currently, AutoHCK can't run clients without the studio. rtoolsHCK also can't establish a connection without the studio because the studio is a proxy to clients.

YanVugenfirer (approved):

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: That behavior and the reasoning behind is explained in the commit message. I hope it makes sense, but we can always make an additional change if it turns out otherwise.

> After this change, the network test bridge will be always created even if network tests are not performed, but the overhead should be negligible, considering the cost of the control bridge is ~torelated~ tolerated.

PR #150, opened 2022-08-18
Create own network namespace

akihikodaki:
Although I said the command invocation will be `bin/ns bin/auto_hck` in an internal meeting, ~~I decided to make it work just by `bin/auto_hck`~~. `bin/ns` still exists for manual testing.

Don't forget to install slirp4netns before testing.

### DIFF
--- a/bin/ns
+++ b/bin/ns
@@ -1,0 +1,48 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'bundler/setup'
+require 'English'
+require 'fileutils'
+require 'tmpdir'
+
+e_read, e_write = IO.pipe
+r_read, r_write = IO.pipe
+e_read.close_on_exec = false
+r_write.close_on_exec = false
+parent = Process.pid
+a_dir = Dir.mktmpdir('autohck')
+a = File.join(a_dir, 'sock')
+
+fork do
+  e_write.close
+  r_read.close
+  Process.setpgid 0, 0
+  e_read.read 1
+
+  slirp = spawn('slirp4netns', '-r', r_write.fileno.to_s,
+                '-a', a, parent.to_s, 'tap_host')
+  begin
+    r_write.close
+    e_read.read 1
+  ensure
+    Process.kill 9, slirp
+    Process.wait
+
+    warn "slirp4nents: unexpected exit status #{$CHILD_STATUS.exitstatus}" if $CHILD_STATUS.termsig != 9
+  end
+ensure
+  e_read.close
+  FileUtils.remove_entry_secure a_dir
+end
+
+e_read.close
+r_write.close
+e_write.close_on_exec = false
+r_read.close_on_exec = false
+
+ENV['AUTOHCK_SLIRP'] = a
+
+exec 'unshare', '--user', '--map-root-user', '--net', '--mount',
+     File.join(__dir__, 'ns_unshared'), e_write.fileno.to_s, r_read.fileno.to_s,
+     *(ARGV.empty? ? ['sh'] : ARGV)
--- a/bin/ns_unshared
+++ b/bin/ns_unshared
@@ -1,0 +1,46 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'English'
+require 'fileutils'
+
+[
+  '/proc/sys/net/bridge/bridge-nf-call-arptables',
+  '/proc/sys/net/bridge/bridge-nf-call-ip6tables',
+  '/proc/sys/net/bridge/bridge-nf-call-iptables'
+].each do
+  File.write _1, '0'
+rescue Errno::ENOENT
+  # br_netfilter is not loaded
+end
+
+e_write = IO.new(ARGV[0].to_i)
+e_write.write '1'
+e_write.flush
+
+r_read = IO.new(ARGV[1].to_i)
+r = r_read.read(1)
+r_read.close
+
+return if r.nil?
+
+[
+  ['mount', '--bind', File.join(__dir__, '..', 'resolv.conf'), '/etc/resolv.conf'],
+  %w[ip link set tap_host up],
+  %w[ip link add br_world type bridge],
+  %w[ip addr add 10.0.2.100/24 dev br_world],
+  %w[ip link set tap_host master br_world],
+  %w[ip link set br_world up],
+  %w[ip route add default via 10.0.2.2],
+  %w[ip link add br_ctrl type bridge],
+  %w[ip link set br_ctrl up],
+  %w[ip link add br_test type bridge],
+  %w[ip link set br_test up]
+].each do
+  Process.wait spawn(*_1)
+  raise $CHILD_STATUS.to_s unless $CHILD_STATUS.success?
+rescue StandardError
+  raise "#{_1}: command failed"
+end
+
+exec(*ARGV[2..])
--- a/lib/setupmanagers/qemuhck/network_manager.rb
+++ b/lib/setupmanagers/qemuhck/network_manager.rb
@@ -38,6 +38,19 @@ module AutoHCK
         Json.read_json(device_json, @logger)
       end
 
+      def read_world_ip(device_name, bridge_name, qemu_replacement_list = {})
+        device = read_device(device_name)
+        type_config = @config['devices']['world']
+        replacement_list = device_replacement_list('world', device, type_config, qemu_replacement_list)
+        mac = replace_string_recursive(type_config['mac'], replacement_list)
+
+        line = File.readlines('/proc/net/arp')[1..].map(&:split).find do
+          _1[3] == mac && _1[5] == bridge_name
+        end
+
+        line&.first
+      end
+
       def create_bridge_commands(bridge, tx_queue_len = 0)
         cmd = [
           "brctl show #{bridge} 1>/dev/null 2>&1 || brctl addbr #{bridge}",
@@ -74,7 +87,7 @@ module AutoHCK
         addr.nil? ? '' : ",addr=#{addr}"
       end
 
-      def device_replacement_list(type, device_info, device_options, device_config, qemu_replacement_list)
+      def device_replacement_list(type, device_info, device_config, qemu_replacement_list)
         replacement_list = {
           '@net_if_name@' => device_config['ifname'],
           '@net_up_script@' => "@workspace@/#{type}_ifup_@run_id@.sh",
@@ -82,18 +95,19 @@ module AutoHCK
           '@net_addr@' => net_addr_cmd(device_config['address']),
           '@bus_name@' => device_config['bus_name'],
           '@device_id@' => format('%02x', @dev_id)
-        }.merge(device_options)
+        }
 
         qemu_replacement_list.merge(device_info['define_variables'].to_h).merge(replacement_list)
       end
 
-      def device_info(type, device_name, device_options, qemu_replacement_list)
+      def device_command_info(type, device_name, command_options, qemu_replacement_list)
         @dev_id += 1
 
         device = read_device(device_name)
         type_config = @config['devices'][type]
 
-        replacement_list = device_replacement_list(type, device, device_options, type_config, qemu_replacement_list)
+        replacement_list = device_replacement_list(type, device, type_config, qemu_replacement_list)
+        replacement_list.merge! command_options
         device_command = replace_string_recursive(device['command_line'].join(' '), replacement_list)
 
         @logger.debug("Device #{device_name} used as #{type} device")
@@ -122,7 +136,7 @@ module AutoHCK
           '@netdev_options@' => netdev_options
         }
 
-        cmd, replacement_list = device_info(type, device_name, options, qemu_replacement_list)
+        cmd, replacement_list = device_command_info(type, device_name, options, qemu_replacement_list)
         create_net_up_script(replacement_list.merge({ '@bridge_name@' => @control_bridge }))
 
         cmd
@@ -143,7 +157,7 @@ module AutoHCK
           '@netdev_options@' => netdev_options
         }
 
-        cmd, replacement_list = device_info(type, device_name, options, qemu_replacement_list)
+        cmd, replacement_list = device_command_info(type, device_name, options, qemu_replacement_list)
         create_net_up_script(replacement_list.merge({ '@bridge_name@' => bridge_name }))
 
         cmd
@@ -160,7 +174,7 @@ module AutoHCK
           '@netdev_options@' => netdev_options
         }
 
-        cmd, replacement_list = device_info(type, device_name, options, qemu_replacement_list)
+        cmd, replacement_list = device_command_info(type, device_name, options, qemu_replacement_list)
         create_net_up_script(replacement_list.merge({ '@bridge_name@' => @test_bridge }))
 
         cmd
@@ -183,7 +197,7 @@ module AutoHCK
           '@netdev_options@' => netdev_options
         }
 
-        cmd, = device_info(type, device_name, options, qemu_replacement_list)
+        cmd, = device_command_info(type, device_name, options, qemu_replacement_list)
 
         cmd
       end
--- a/lib/setupmanagers/qemuhck/network_manager.rb
+++ b/lib/setupmanagers/qemuhck/network_manager.rb
@@ -22,9 +22,6 @@ module AutoHCK
         @logger = logger
         @dev_id = 0
 
-        @control_bridge = "br_ctrl_#{id}"
-        @test_bridge = "br_test_#{id}"
-
         @config = Json.read_json(CONFIG_JSON, @logger)
       end
 
@@ -38,49 +35,25 @@ module AutoHCK
         Json.read_json(device_json, @logger)
       end
 
-      def read_world_ip(device_name, bridge_name, qemu_replacement_list = {})
+      def find_world_ip(*args)
+        UDPSocket.open do |socket|
+          (15..30).map { socket.send('', 0, "10.0.2.#{_1}", 9) }
+        end
+
+        read_world_ip(*args)
+      end
+
+      def read_world_ip(device_name, qemu_replacement_list = {})
         device = read_device(device_name)
         type_config = @config['devices']['world']
         replacement_list = device_replacement_list('world', device, type_config, qemu_replacement_list)
         mac = replace_string_recursive(type_config['mac'], replacement_list)
 
         line = File.readlines('/proc/net/arp')[1..].map(&:split).find do
-          _1[3] == mac && _1[5] == bridge_name
+          _1[3] == mac && _1[5] == 'br_world'
         end
 
         line&.first
-      end
-
-      def create_bridge_commands(bridge, tx_queue_len = 0)
-        cmd = [
-          "brctl show #{bridge} 1>/dev/null 2>&1 || brctl addbr #{bridge}",
-          "ifconfig #{bridge} up"
-        ]
-
-        cmd << "ifconfig #{bridge} txqueuelen #{tx_queue_len}" if tx_queue_len.positive?
-
-        cmd
-      end
-
-      def remove_bridge_commands(bridge)
-        [
-          "ifconfig #{bridge} down",
-          "brctl delbr #{bridge}"
-        ]
-      end
-
-      def disable_bridge_nf_commands
-        [
-          'sysctl -e net.bridge.bridge-nf-call-arptables=0',
-          'sysctl -e net.bridge.bridge-nf-call-ip6tables=0',
-          'sysctl -e net.bridge.bridge-nf-call-iptables=0'
-        ]
-      end
-
-      def disable_bridge_nf
-        @logger.info('Disabling bridge-netfilter')
-
-        disable_bridge_nf_commands
       end
 
       def net_addr_cmd(addr)
@@ -137,16 +110,12 @@ module AutoHCK
         }
 
         cmd, replacement_list = device_command_info(type, device_name, options, qemu_replacement_list)
-        create_net_up_script(replacement_list.merge({ '@bridge_name@' => @control_bridge }))
+        create_net_up_script(replacement_list.merge({ '@bridge_name@' => 'br_ctrl' }))
 
         cmd
       end
 
-      def control_bridge_command
-        [create_bridge_commands(@control_bridge), remove_bridge_commands(@control_bridge)]
-      end
-
-      def world_device_command(device_name, bridge_name, qemu_replacement_list = {})
+      def world_device_command(device_name, qemu_replacement_list = {})
         type = __method__.to_s.split('_').first
 
         netdev_options = ',vhost=@vhost_value@,script=@net_up_script@,downscript=no'
@@ -158,7 +127,7 @@ module AutoHCK
         }
 
         cmd, replacement_list = device_command_info(type, device_name, options, qemu_replacement_list)
-        create_net_up_script(replacement_list.merge({ '@bridge_name@' => bridge_name }))
+        create_net_up_script(replacement_list.merge({ '@bridge_name@' => 'br_world' }))
 
         cmd
       end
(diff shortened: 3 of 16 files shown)


lib/engines/hcktest/tests.rb
```diff
@@ -309,7 +308,6 @@ def new_done
     def apply_filters
       @logger.info('Applying filters on finished tests')
       @tools.apply_project_filters(@tag)
-      sleep APPLYING_FILTERS_INTERVAL
```
kostyanf14: looks like this is not related to DHCPD. Please remove or move to separate commit if this is needed.

lib/engines/hcktest/tools.rb
```diff
@@ -22,8 +22,12 @@ def initialize(project, ip_addr, clients)
               logger: @logger,
               outp_dir: project.workspace_path,
               l_script_file: @config['toolshck_path'])
+
+      ObjectSpace.define_finalizer(self, FINALIZER)
```
kostyanf14: Why do we need some finalizer there?

lib/setupmanagers/qemuhck/dhcpd.rb
```diff
@@ -0,0 +1,79 @@
...
+      dump_conf conf_path
+      FileUtils.touch leases_path
+
+      if dump_only
+        IO.write File.join(@workspace_path, 'dhcpd_manual.sh'), <<~SH
+          #!/usr/bin/env bash
+          dhcpd -f -d -cf #{conf_path} -lf #{leases_path} --no-pid
+        SH
+      else
+        @pid = spawn('dhcpd', '-f', '-d', '-cf', conf_path, '-lf', leases_path, '--no-pid')
```
kostyanf14, line 28: What if `dhcpd` will fail to start? Will auto_hck fail too? 

What if `dhcpd` will fail during the HCK test run? Will auto_hck restart `dhcpd`?

lib/setupmanagers/qemuhck/dhcpd.rb
```diff
@@ -0,0 +1,79 @@
...
+
+          ddns-update-style none;
+
+          subnet #{@ip_segment}0 netmask 255.255.255.0 {
+
+          option routers #{@ip_segment}1;
+          range #{@ip_segment}200 #{@ip_segment}254;
+        CONF
+
+        (2..99).each do |i|
```
kostyanf14, line 56: I think this is bad to hardcode `2..99`. We have no way when DHCP server was installed as a separate application, but now we run it as part of AutoHCK. So, I think we can get this range from config.

Also if we run a separate `dhcpd` for each `auto_hck` we can create config only for the current ID.

lib/setupmanagers/qemuhck/dhcpd.rb
```diff
@@ -0,0 +1,79 @@
...
+          f.write <<~CONF
+            host qemuhck#{i2} {
+              hardware ethernet 56:00:#{i2}:00:dd:dd;
+              fixed-address #{@ip_segment}#{i};
+            }
+
+            host virthck#{i2} {
+              hardware ethernet 56:00:#{i2}:00:#{i2}:dd;
+              fixed-address #{@ip_segment}#{i};
+            }
```
kostyanf14, line 68: AutoHCK does not have virthck support now. This is present in the DHCP configuration because a user can reuse the same bridge for VirtHCK. So just remove this part.

bin/auto_hck
```diff
@@ -11,6 +11,8 @@ begin
   require './lib/project'
   require './lib/trap'
 
+  exec File.join(__dir__, 'ns'), __FILE__, *ARGV unless ENV.key?('AUTOHCK_SLIRP')
```
kostyanf14, line 14: Is this safe to start something via exec?

bin/ns_unshared
```diff
@@ -0,0 +1,32 @@
...
+
+e_write = IO.new(ARGV[0].to_i)
+r_read = IO.new(ARGV[1].to_i)
+e_write.write('1')
+e_write.flush
+r_read.read(1)
+r_read.close
+ENV['AUTOHCK_SLIRP'] = File.join(ARGV[2], 'sock')
+resolv = File.join(ARGV[2], 'resolv.conf')
+IO.write resolv, 'nameserver 10.0.2.3'
```
kostyanf14, line 20: What is this IP? Should it be hard coded?

kostyanf14 (changes requested):
Quick review. I will look detailed and test this later.

lib/engines/hcktest/tests.rb
akihikodaki: I removed this. It is an irrelevant change.

lib/engines/hcktest/tools.rb
akihikodaki: Removed this too. It was for debugging.

lib/setupmanagers/qemuhck/dhcpd.rb
```diff
@@ -0,0 +1,79 @@
...
+      Process.kill(:KILL, @pid)
+      Process.wait(@pid)
+    end
+
+    private
+
+    # rubocop:disable Metrics/MethodLength
+    def dump_conf(conf_path)
+      File.open(conf_path, 'w') do |f|
+        f.write <<~CONF
```
kostyanf14, line 44: Also, you can use ruby/erb and move this config to a separate file. Just an idea.

lib/setupmanagers/qemuhck/dhcpd.rb
akihikodaki: No for both. I don't have an idea how to handle those cases for now.

bin/auto_hck
akihikodaki: I think so. We already have multiple `spawn` calls, which should include `exec`.

bin/ns_unshared
akihikodaki: It is defined by slirp4netns. See: https://github.com/rootless-containers/slirp4netns/blob/master/slirp4netns.1.md#description

YanVugenfirer:
Looks Ok from my side. But I would be happy to see a ran of network driver (NetKVM\virtio-net) with all the latest changes.

akihikodaki:
a0cdb0fbc36926c1afa9494b1f2b5e289560542e:

- Reworked to use DHCP server functionality of slirp4netns.
- Replaced `brctl` with `ip`.
- Added code to close unused files in the forked process of `bin/ns`.
- Added code to create a process group for the forked process of `bin/ns` to limit signals sent from the shell

kostyanf14:
@akihikodaki 

I tried to start CI on Ubuntu 20.04 and failed to do this. Command line `rvmsudo ./bin/auto_hck test --drivers NetKVM --platform Win2022x64 --driver-path /mnt/NetKVM/2k22/amd64`. I got the following errors:

if `id_gen.db` exist, I got `I, [2022-08-22T08:08:18.162024 #2377]  INFO -- : No available ID`, but DB is clean, so all ID should be available.
if `id_gen.db` is missing, I got a runtime error:
```
/usr/local/rvm/gems/ruby-3.0.4/gems/sqlite3-1.4.2/lib/sqlite3/database.rb:89:in `open_v2': unable to open database file (SQLite3::CantOpenException)
        from /usr/local/rvm/gems/ruby-3.0.4/gems/sqlite3-1.4.2/lib/sqlite3/database.rb:89:in `initialize'
        from /home/user/repos/AutoHCK/lib/auxiliary/id_gen.rb:13:in `new'
        from /home/user/repos/AutoHCK/lib/auxiliary/id_gen.rb:13:in `initialize'
        from /home/user/repos/AutoHCK/lib/project.rb:121:in `new'
        from /home/user/repos/AutoHCK/lib/project.rb:121:in `assign_id'
        from /home/user/repos/AutoHCK/lib/project.rb:35:in `initialize'
        from ./bin/auto_hck:30:in `new'
        from ./bin/auto_hck:30:in `<module:AutoHCK>'
        from ./bin/auto_hck:17:in `<main>'
```

akihikodaki:
The below is the description of this change for an internal meeting.

# Concepts

## Network bridge

A network _bridge_ is basically a virtual L2 switch. The _taps_ in a bridge are connected with each other, and the bridge passes IP packets among them.
A bridge is globally scoped, uniquely identified with string identifiers, and it persists until the system reboots or it is explicitly removed. This nature of bridges makes managing them difficult.
As bridges has no concept of ownership, managing bridges requires CAP_NET_ADMIN capability (i.e. sudo everywhere).

## Namespace

_Namespace_ is the fundamental idea of this change. It creates an isolated environment for global resources. In this case, network namespace plays the critical role to create an isolated environment for bridges. Container runtimes like Docker are based on complex combinations of namespace and other constructs like cgroups, but it is also possible to use some namespace construct selectively and to create a semi-isolated environment.

Namespace is hierarchical. The system starts with the root namespace. Creating a namespace divides the world into the outer namespace and the inner namespace. It is also possible to nest namespaces.

Creating a namespace does not require a privilege.

# Using AutoHCK with namespaces

Run `bin/ns` to create namespaces, configure the inner network for AutoHCK and execute `sh` as root in them. You can also run e.g. `bin/ns bash` to get Bash in the inner namespaces.

`bin/auto_hck` automatically runs `bin/ns` if it is executed from namespaces not created by `bin/ns` so prepending `bin/ns` to `bin/auto_hck` is not required.

`bin/ns` is still useful to manually run tests, for example.

# Tools

These are the main tools this change uses.

## `unshare`

`unshare` creates namespaces and executes a command in the namespace. This change creates the following namespaces:
- `user` namespace as it is required by `net` namespace. With `user` namespace, it is possible to run a command as a root in the inner namespace even if you don't have the capability in the outer namespace.
- `net` namespace to create bridges in an isolated environment.
- `mount` namespace to override `/etc/resolv.conf` to point to slirp4netns.

## `slirp4netns`

`slirp4netns` is a daemon for network namespace. The below is the roles of slirp4netns.

- Gateway to provide connectivity from the inner namespace to the outer namespace. An alternative way to provide such connectivity is to have `veth`s, virtual L2 ports that can cross namespace boundaries. veth more efficient as it is a kernel-space mechanism, but managing veth for the outer namespace has the same difficulty with bridges. 
- Forwarder from the outer namespace to the inner namespace. slirp4netns listens to ports in the outer namespace, and forwards packets to them to the inner namespace. The forwarding configuration can be altered at runtime with a Unix domain socket slirp4netns listens to.
- DNS server. Some system has an internal DNS server to e.g. forward unencrypted queries via DNS-over-TLS, but the inner namespace cannot connect to the server by default. In the case, slirp4netns can forward queries to the outer namespace.
- DHCP server.

## `ip`

`ip` is a generic network management command. Before this change, some specific tools, namely `brctl` and `ifconfig` were used to manage different network constructs, but they are now managed by `ip` since `ip` is necessary to create tap devices anyway.

# Diagram

The below describes the sequence to start installing a Windows image and aborting it.

```mermaid
sequenceDiagram
    participant Microsoft
    participant User
    participant bin/ns
    participant Forked bin/ns
    participant slirp4netns
    activate User
    User ->> bin/ns: Spawn bin/ns
    deactivate User
    activate bin/ns
    bin/ns ->> bin/ns: Create a temporary directory
    bin/ns ->> Forked bin/ns: Fork
    bin/ns ->> bin/ns: Execute `unshare (snip) bin/ns_unshared`
    bin/ns ->> bin/ns: Set up bridges and taps
    bin/ns ->> Forked bin/ns: Write pipe `e`
    deactivate bin/ns
    activate Forked bin/ns
    Forked bin/ns ->> slirp4netns: Spawn slirp4netns
    deactivate Forked bin/ns
    activate slirp4netns
    slirp4netns ->> bin/ns: Write pipe `r`
    deactivate slirp4netns
    activate bin/ns
    bin/ns ->> bin/ns: Execute `bin/auto_hck`
    bin/ns ->> slirp4netns: Request to start port forwarding
    User ->> slirp4netns: Connect for VNC
    slirp4netns ->> bin/ns: Connect for VNC
    bin/ns ->> slirp4netns: Request FOD
    slirp4netns ->> Microsoft: Request FOD
    bin/ns ->> slirp4netns: Request to stop port forwarding
    User ->> bin/ns: Send SIGINT twice
    bin/ns ->> Forked bin/ns: Exit, closing pipe `e`
    deactivate bin/ns
    activate Forked bin/ns
    Forked bin/ns ->> slirp4netns: SIGKILL
    Forked bin/ns ->> Forked bin/ns: Remove the temporary directory
    deactivate Forked bin/ns
```

akihikodaki:
08084ef971e3f37bf29702272ff097ff46e31c12:

- Let slirp4nets create own tap
- Name the tap for slirp4netns `tap_host`
- Describe the dependencies of commands in `bin/ns_unshare`

akihikodaki:
@kostyanf14 I could reproduce the issue with `sudo`. The below fixed for me:

```
chmod a+x /home/person
chown -R root.root /home/person/HCK-CI
```

So please make sure all of the parent directories have the executable bits set and the installation is owned by root.

KVM did work without root on my installation of Ubuntu 20.04. `ls -l /dev/kvm` says:

```
$ ls -l /dev/kvm
crw-rw----+ 1 root kvm 10, 232 Aug 22 23:31 /dev/kvm
```

akihikodaki:
20ab14173315c000ee783b462b026650b5fbf963: Use a static `resolv.conf` instead of dynamically creating one, simplifying the code.

akihikodaki:
2dcd27c50918cc0b2aca0204a9efeaeff3231b64: Replaced net-ping with `UDPSocket`. net-ping catches `SystemExit` and makes aborting impossible. `UDPSocket` is sufficient for this because we just want to update ARP table here and no response is needed.

lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -38,6 +38,19 @@ def read_device(device)
         Json.read_json(device_json, @logger)
       end
 
+      def read_world_ip(device_name, bridge_name, qemu_replacement_list = {})
+        device = read_device(device_name)
+        type_config = @config['devices']['world']
+        replacement_list = device_replacement_list('world', device, type_config, qemu_replacement_list)
+        mac = replace_string_recursive(type_config['mac'], replacement_list)
+
+        line = File.readlines('/proc/net/arp')[1..].map(&:split).find do
```
kostyanf14, line 52: The idea looks good, but I am not sure about the stability. When I ran VM and then stopped it (without network name space), VM MAC still was present in the arp table for a long time.

bin/ns
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env ruby
```
kostyanf14, line 1: What about logs of these scripts (`bin/ns`, `bin/ns_unshared`)? Do we need them for debugging? Currently, it presents only in the console.

bin/auto_hck
kostyanf14, line 14: From docs: `exec(*args) - Replaces the current process by running the given external command`
We need network namespaces and slirp only for Linux + QEMU. We should have access to the real network in the case of Physical HCK and for Windows (plans). 

So, we need a way to run auto_hck without network namespace.

kostyanf14 (changes requested):

kostyanf14:
@akihikodaki 

This fix https://github.com/HCK-CI/AutoHCK/pull/150#issuecomment-1222449125 works for me.

kostyanf14:
@akihikodaki

I got one more problem. Steps to reproduce:
1. run ./bin/auto_hck
2. wait until 3 VMs starts
3. press CTRL+C twice.

Currently, I got some errors from QEMU and auto_hck hangs on. Logs:

```
I, [2022-09-01T08:58:53.371738 #2613]  INFO -- : Configuring client CL1...                                                                                                                       [643/1269]
I, [2022-09-01T08:58:53.371910 #2613]  INFO -- : Moving CL1 to pool                                                                                                                                        
^CW, [2022-09-01T08:58:55.492678 #2613]  WARN -- : SIGINT(1) received, aborting if another SIGINT is received in the span of the next one second                                                           
W, [2022-09-01T08:58:55.559954 #2613]  WARN -- server/run_log_fetcher: Errno::ECONNREFUSED: Connection refused - Connection refused - connect(2) for "10.0.2.15" port 5985 (10.0.2.15:5985)                
^CW, [2022-09-01T08:58:55.680587 #2613]  WARN -- : SIGINT(2) received, aborting...                                                                                                                         
D, [2022-09-01T08:58:55.680843 #2613] DEBUG -- : Closing AutoHCK project                                                                                                                                   
W, [2022-09-01T08:58:55.959503 #2613]  WARN -- : Error dump:                                                                                                                                               
   -- qemu-system-x86_64: -netdev tap,id=cs_r0002_c02,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/control_ifup_0002.sh,downscript=no: warni
ng: tap: open vhost char device failed: Permission denied                                                                                                                                                  
   -- qemu-system-x86_64: -netdev tap,id=ts_r0002_c02,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/test_ifup_0002.sh,downscript=no: warning:
 tap: open vhost char device failed: Permission denied                                                                                                                                                     
   -- qemu-system-x86_64: terminating on signal 2                                                                                                                                                          
W, [2022-09-01T08:58:55.979203 #2613]  WARN -- : Error dump:                                                                                                                                               
   -- qemu-system-x86_64: -netdev tap,id=cs_r0002_c00,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/control_ifup_0002.sh,downscript=no: warni
ng: tap: open vhost char device failed: Permission denied                                                                                                                                                  
   -- qemu-system-x86_64: -netdev tap,id=ws_r0002_c00,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/world_ifup_0002.sh,downscript=no: warning
: tap: open vhost char device failed: Permission denied                                                                                                                                                    
   -- qemu-system-x86_64: terminating on signal 2                                                                                                                                                          
W, [2022-09-01T08:58:55.979402 #2613]  WARN -- : Error dump:                                                                                                                                               
   -- qemu-system-x86_64: -netdev tap,id=cs_r0002_c01,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/control_ifup_0002.sh,downscript=no: warni
ng: tap: open vhost char device failed: Permission denied                                                                                                                                                  
   -- qemu-system-x86_64: -netdev tap,id=ts_r0002_c01,vhost=on,script=/storage/data/user/auto_hck/workspace/hcktest/qemuhck/NetKVM-Win2022x64/2022_09_01_08_53_40/test_ifup_0002.sh,downscript=no: warning:
 tap: open vhost char device failed: Permission denied                                                                                                                                                     
   -- qemu-system-x86_64: terminating on signal 2                                                                                                                                                          
apifd event                                                                                                                                                                                                
api_handler: got request: {"execute":"remove_hostfwd","arguments":{"id":3}}                                                                                                                                
apifd event                                                                                                                                                                                                
api_handler: got request: {"execute":"remove_hostfwd","arguments":{"id":4}}                                                                                                                                
I, [2022-09-01T08:58:56.679389 #2613]  INFO -- : QemuMachine0002_CL01 is not alive                                                                                                                         
apifd event                                                                                                                                                                                                
api_handler: got request: {"execute":"remove_hostfwd","arguments":{"id":5}}                                                                                                                                
apifd event                                                                                                                                                                                                
api_handler: got request: {"execute":"remove_hostfwd","arguments":{"id":6}}                                                                                                                                
I, [2022-09-01T08:58:56.679816 #2613]  INFO -- : QemuMachine0002_CL02 is not alive
I, [2022-09-01T08:58:56.679869 #2613]  INFO -- : Aborting HLK Studio
```

I wait about 10 minutes and auto_hck still was at this stage.

lib/setupmanagers/qemuhck/network_manager.rb
akihikodaki: It shouldn't be a problem since the ARP table is isolated by network namespace.

bin/ns
akihikodaki: It is unlikely that they will be needed for debugging HLK tests. `bin/ns` and `bin/ns_unshared` only concerns about initialization, cleanup, and forwarding between the outer namespace (for Internet and operator access) and the inner namespace.
The network constructs the guest will operate on are entirely contained in a network namespace so we don't have to concern about `bin/ns` and `bin/ns_unshared` once the namespace is constructed.

If you need to debug a setup issue (e.g. you tried `bin/ns` on a new system setup but it has namespace disabled), you may need the logs, but the console output should be enough in the case.

bin/auto_hck
akihikodaki: I just dropped the line with commit 65b71aecd7e1a3bd676a1c5cff353bbd331fe9ab. Now you always need to prefix the command line with `bin/ns` if you run QEMU.

kostyanf14: ok

akihikodaki:
@kostyanf14 https://github.com/HCK-CI/AutoHCK/pull/152/commits/ae632a0865ad39575c73d5d75bff6e0cb6f68382 should fix the problem. I guess you used sudo before this change. Now if you use AutoHCK without sudo, the shell directly signals AutoHCK. Unlike sudo, the shell signals the process group, resulting QEMU killed before AutoHCK cleans up connections to QEMU.
I also realized monitor commands do not work well for abortion so added another commit: https://github.com/HCK-CI/AutoHCK/pull/152/commits/67521fa13b63cd126fbf2e183bb32001871995e8.

kostyanf14:
@akihikodaki 

>resulting QEMU killed before AutoHCK cleans up connections to QEMU

I think that AutoHCK should be properly terminated.  Currently, we have the same termination order, but AutoHCK successfully exits.

kostyanf14:
@YanVugenfirer @akihikodaki 

HCK tests for NetKVM still running, but two tests are already failed (`NDISTest 6.5 - [2 Machine] - OffloadChecksum` and `NDISTest 6.5 - [2 Machine] - GlitchFreeDevice`). I use NetKVM from VirtIO ISO, so it must be stable.
As I remember to fix `NDISTest 6.5 - [2 Machine] - OffloadChecksum` we add:
```
sysctl -e net.bridge.bridge-nf-call-arptables=0
sysctl -e net.bridge.bridge-nf-call-ip6tables=0
sysctl -e net.bridge.bridge-nf-call-iptables=0
```

So, the network namespace still enables bridge net filters.

YanVugenfirer:
> @YanVugenfirer @akihikodaki
> 
> HCK tests for NetKVM still running, but two tests are already failed (`NDISTest 6.5 - [2 Machine] - OffloadChecksum` and `NDISTest 6.5 - [2 Machine] - GlitchFreeDevice`). I use NetKVM from VirtIO ISO, so it must be stable. As I remember to fix `NDISTest 6.5 - [2 Machine] - OffloadChecksum` we add:
> 
> ```
> sysctl -e net.bridge.bridge-nf-call-arptables=0
> sysctl -e net.bridge.bridge-nf-call-ip6tables=0
> sysctl -e net.bridge.bridge-nf-call-iptables=0
> ```
> 
> So, the network namespace still enables bridge net filters.

Regarding - GlitchFreeDevice please check the logs. We had some issues with it if you remember and it was failing statistically or we were getting warning. I would test it on regular deployment just in case.

For OffloadChecksum we need the mentioned commands. So probably they should be executed in the context of namespace, no?

akihikodaki:
@kostyanf14 @YanVugenfirer Added sysctl changes with commit 27ee5c7439530f13dfe2eb0ad8d484b2ed1097bd. Please try again with it.

kostyanf14:
@akihikodaki Please rebase this PR to master

akihikodaki:
cbfa544a3c6a760ff6e53c43216faf242c664aa7: rebased to bf11341b56ddbb9cd9ce6072d7b1d2b6998112d3

YanVugenfirer (approved):

akihikodaki:
# Checklist after merging

- [ ] Update AutoHCK-Installer
- [ ] Bump supported Fedora versions to 34, 35 and 36
- [ ] Note required slirp4netns version (v0.4.0)

kostyanf14 (approved):
LGTM. 

@akihikodaki Please add commit with AutoHCK version bump. We have a big change, so I think to bump to 0.10.0

akihikodaki:
@kostyanf14 Bumped the version with commit 86c32312531b10c7d65b0e719df94307f566697a.

PR #154, opened 2022-09-19
Fix crash in check_test_queued_time

kostyanf14:


akihikodaki:
I don't think failing to filter, which is fixed with commit 4ebd146ca35d044ebc764d442b943623ae215435, is contributing to crash in `check_test_queued_time`. Is the commit included in this pull request just for convenience, or is it actually necessary to fix the crash?

PR #155, opened 2022-09-23
engines: Abort each run

akihikodaki:
Not tested.

### DIFF
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -617,7 +617,6 @@ module AutoHCK
 
       remove_hostfwd
       vm_abort
-      @nm.close
       run_post_stop_commands
     end
   end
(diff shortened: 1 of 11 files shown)


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -54,8 +54,6 @@ def initialize(options)
 
     def define_local_variables
       @devices_list = []
-      @pre_start_commands = []
```
kostyanf14, line 57: Why did you remove this? We have some devices such as ivshmem/virtio-fs where we MUST run the daemon on the host before tests and kill it after it.

kostyanf14 (commented):
I am testing these patches. I will comment more later.

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: I missed their usage. Commit e82eedb3dfaef22b24a28a4972a0bfe6b4dd8938 restores them.

kostyanf14: For example: https://github.com/HCK-CI/AutoHCK/blob/master/lib/setupmanagers/qemuhck/devices/ivshmem-doorbell.json

YanVugenfirer (approved):

PR #156, opened 2022-11-03
Bump RuboCop version (1.38.0) 

kostyanf14:


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -393,7 +393,7 @@ def dump_config
     end
 
     def create_run_script(file_name, file_content)
-      File.open(file_name, 'w') { |f| f.write(file_content) }
+      File.write(file_name, file_content)
```
YanVugenfirer, line 396: Is open somewhere else, or they change the write function in Ruby?

akihikodaki: `File` class has a class method version of `write`, which opens, writes and closes file.

lib/config/sentry.rb
```diff
@@ -31,9 +31,9 @@ module AutoHCK
     # of transactions for performance monitoring.
     config.traces_sample_rate = 1.0
 
-    unless ENV['SENTRY_TRANSPORT_SSL_CA'].nil?
-      if File.exist?(ENV['SENTRY_TRANSPORT_SSL_CA'])
-        config.transport.ssl_ca_file = ENV['SENTRY_TRANSPORT_SSL_CA']
+    unless ENV.fetch('SENTRY_TRANSPORT_SSL_CA', nil).nil?
```
akihikodaki, line 34: `ENV.key?` can be used if you just want to check its existence. I prefer to store the result of `ENV.fetch('SENTRY_TRANSPORT_SSL_CA', nil)` to a variable and use it for both of the existence check and later use in Sentry configuration though.

lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -49,8 +49,8 @@ def read_world_ip(device_name, qemu_replacement_list = {})
         replacement_list = device_replacement_list('world', device, type_config, qemu_replacement_list)
         mac = replace_string_recursive(type_config['mac'], replacement_list)
 
-        line = File.readlines('/proc/net/arp')[1..].map(&:split).find do
-          _1[3] == mac && _1[5] == 'br_world'
+        line = File.readlines('/proc/net/arp')[1..].map(&:split).find do |arp_line|
```
akihikodaki, line 52: Distinguishing two variables by naming `line` and `arp_line` sounds too ambiguous. Perhaps we may use rename `line` to `found` and `arp_line` to `candidate`. You may not prefer `candidate` since it is a bit verbose though.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -420,23 +420,23 @@ def run_vm
...
+                                               'host_port' => port,
+                                               'guest_port' => port
                                              }
                                            })
       end
     end
 
     def remove_hostfwd
-      @hostfwds.each do
+      @hostfwds.each do |fd|
```
akihikodaki, line 436: `fd` sounds like a file descriptor, but it is not.

lib/trap.rb
```diff
@@ -32,8 +32,8 @@ def self.normal_trap(signal)
         perform_trap(signal)
       else
         @sig_timestamps[signal] = time
-        write_log("SIG#{signal}(1) received, aborting if another SIG#{signal} is"\
-                            ' received in the span of the next one second')
+        write_log("SIG#{signal}(1) received, aborting if another SIG#{signal} is " \
```
akihikodaki, line 35: In general, having a line break in a string literal is not great. It makes difficult to search the message with grep. Such a long message is often bad not only in the source code but actual output. It is usually better to add a line break also in the actual output or shorten the message.

bin/ns_unshared
```diff
@@ -8,10 +8,8 @@ require 'fileutils'
   '/proc/sys/net/bridge/bridge-nf-call-arptables',
   '/proc/sys/net/bridge/bridge-nf-call-ip6tables',
   '/proc/sys/net/bridge/bridge-nf-call-iptables'
-].each do
-  File.write _1, '0'
-rescue Errno::ENOENT
-  # br_netfilter is not loaded
+].each do |file|
+  File.write file, '0' if File.exist? file
```
akihikodaki, line 12: Replacing `Errno::ENOENT` rescue with `File.exist?` check is not good as it is possible to remove the file after `File.exist?` check while such a failure is not possible with `Errno::ENOENT` rescue. Though such a failure is unlikely to happen in this particular code, the sequence of file opening after checking should be considered as a common anti-pattern in UNIX programming and avoided.

lib/setupmanagers/hckclient.rb
```diff
@@ -198,8 +198,8 @@ def synchronize(exit: false)
     end
 
     def not_ready?
-      @studio.list_pools.detect { |pool| pool['name'].eql?(@pool) }['machines']\
-             .detect { |machine| machine['name'].eql?(@name) }['state']\
+      @studio.list_pools.detect { |pool| pool['name'].eql?(@pool) }['machines'] \
```
akihikodaki, line 201: I don't think the backslashes are necessary at first place.

lib/auxiliary/id_gen.rb
```diff
@@ -16,9 +16,9 @@ def initialize(range, timeout)
 
     def load_data
       ids = Array.new(@range.last + 1 - @range.first, true)
-      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds('\
-                  'IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,'\
-                  'Time TIMESTAMP NOT NULL)'
+      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds(' \
```
akihikodaki, line 19: Perhaps it is better to use heredoc instead?

akihikodaki (changes requested):
Good to see that the new version of RuboCop finds more room of improvements. I added comments where I came up with alternatives to fix new offenses.

bin/ns_unshared
kostyanf14: The latest rubocop ask me to remove `rescue`. I will think about how to rewrite this part.

akihikodaki: Weird. It seems not complained about an empty `rescue` in `lib/setupmanagers/qemuhck/qemu_machine.rb`.

akihikodaki: It seems that an offense of `Style/NumberedParameters` somewhat triggers `Lint/SuppressedException`. Fixing `Style/NumberedParameters` also suppresses `Lint/SuppressedException`. Likely a bug of RuboCop.

akihikodaki: Opened https://github.com/rubocop/rubocop/pull/11151.

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: Moved to https://github.com/HCK-CI/AutoHCK/pull/158

lib/auxiliary/id_gen.rb
kostyanf14: Moved to https://github.com/HCK-CI/AutoHCK/pull/161

lib/config/sentry.rb
kostyanf14: Moved to https://github.com/HCK-CI/AutoHCK/pull/160

lib/setupmanagers/qemuhck/network_manager.rb
kostyanf14: Moved to https://github.com/HCK-CI/AutoHCK/pull/164

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: Fixed in https://github.com/HCK-CI/AutoHCK/pull/164

akihikodaki (approved):

YanVugenfirer (approved):

PR #161, opened 2022-11-07
Use heredoc for SQL

kostyanf14:
Signed-off-by: Kostiantyn Kostiuk <konstantin@daynix.com>

### DIFF
--- a/lib/auxiliary/id_gen.rb
+++ b/lib/auxiliary/id_gen.rb
@@ -16,9 +16,12 @@ module AutoHCK
 
     def load_data
       ids = Array.new(@range.last + 1 - @range.first, true)
-      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds('\
-                  'IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,'\
-                  'Time TIMESTAMP NOT NULL)'
+      @conn.execute <<~SQL
+        CREATE TABLE IF NOT EXISTS ActiveIds(
+          IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,
+          Time TIMESTAMP NOT NULL
+        )
+      SQL
       fetch_data(ids)
       ids
     rescue SQLite3::Exception
@@ -74,8 +77,10 @@ module AutoHCK
     end
 
     def kill_if_alive(now)
-      res = @conn.execute 'SELECT PID FROM ActiveIds WHERE'\
-                          "(#{now} - Time) > #{@threshold}"
+      res = @conn.execute <<~SQL
+        SELECT PID FROM ActiveIds
+        WHERE (#{now} - Time) > #{@threshold}
+      SQL
       res.each do |row|
         pid = row[0].to_i
         check_cmd = "sudo ps -ef | grep #{pid} | grep auto_hck | grep -v grep"


lib/auxiliary/id_gen.rb
```diff
@@ -16,9 +16,12 @@ def initialize(range, timeout)
 
     def load_data
       ids = Array.new(@range.last + 1 - @range.first, true)
-      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds('\
-                  'IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,'\
-                  'Time TIMESTAMP NOT NULL)'
+      @conn.execute <<-SQL
```
akihikodaki, line 19: Use `~` instead of `-` ("squiggly" heredoc)

akihikodaki (changes requested):

lib/auxiliary/id_gen.rb
```diff
@@ -16,9 +16,12 @@ def initialize(range, timeout)
 
     def load_data
       ids = Array.new(@range.last + 1 - @range.first, true)
-      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds('\
-                  'IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,'\
-                  'Time TIMESTAMP NOT NULL)'
+      @conn.execute <<~SQL
+          CREATE TABLE IF NOT EXISTS ActiveIds(
```
akihikodaki, line 20: This has four-column indent while everywhere else has two-column indent. (So RuboCop also misses indents in sluggish heredoc? :confounded:)

akihikodaki (changes requested):

lib/auxiliary/id_gen.rb
akihikodaki: What? Then why not in `kill_if_alive`? :confused:

kostyanf14: Everything ok if `-SQL`, but when `~SQL` RuboCop generate offence:
```
lib/auxiliary/id_gen.rb:20:1: C: [Correctable] Layout/HeredocIndentation: Use 2 spaces for indentation in a heredoc
```

akihikodaki: I think I get the problem now. The closing brace is indented with 2 spaces so RuboCop does no longer complain, but the first line still has a wrong indent. It should be fixed so that it will have 2 spaces.

kostyanf14: Thanks for explanation, fixed

lib/auxiliary/id_gen.rb
```diff
@@ -16,9 +16,12 @@ def initialize(range, timeout)
 
     def load_data
       ids = Array.new(@range.last + 1 - @range.first, true)
-      @conn.execute 'CREATE TABLE IF NOT EXISTS ActiveIds('\
-                  'IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,'\
-                  'Time TIMESTAMP NOT NULL)'
+      @conn.execute <<~SQL
+        CREATE TABLE IF NOT EXISTS ActiveIds(
+        IdNumber INT PRIMARY KEY NOT NULL, PID INT NOT NULL,
```
akihikodaki, line 21: I think you should add more indents in braces because you would do so in Ruby while this is in SQL.

akihikodaki (changes requested):

akihikodaki (approved):

PR #181, opened 2023-01-23
Add missing test extra

kostyanf14:
This is a workaround for production to prevent crashes and
ability to collect more logs for debugging.


akihikodaki (commented):
I guess the cause of the crash is `nil` is assigned to `@last_queued_id` at line 303.
https://github.com/HCK-CI/AutoHCK/pull/181/files#diff-50be1d6d012841e6d5183c4a6e3540f497bc92bb49d8567f795e50a4fe91479fR303

PR #208, opened 2023-05-04
Updates for running Win2022 SVVP

kostyanf14:


lib/engines/hckinstall/hckinstall.rb
```diff
@@ -304,11 +325,7 @@ def tag
       "install-#{@project.options.install.platform}"
     end
 
-    def run
-      @logger.debug('HCKInstall: run')
-
-      prepare_setup_scripts_config
-
+    def run_studio_installation
```
akihikodaki, line 328: This name is somewhat confusing. `install_studio` is the one which calls `run_studio` so it can be even said that this name is more appropriate for that method.

I think it's good time to rename the existing `install_studio`, `install_client` and `install_clients` to more descriptive names. In particular, the new names should express that those methods consume ISOs created by e.g. `prepare_studio_iso`. For example, `install_studio` can be renamed to `run_studio_installer` (not using the word "iso" here because it's a bit awkward to say "run ISO") and `prepare_studio_iso` can be renamed to `prepare_studio_installer` accordingly.

After this refactoring, `run_studio_installation` can be named as `install_studio`.

akihikodaki (changes requested):

YanVugenfirer (approved):
Other than @akihikodaki 's comment, approved

akihikodaki (changes requested):
`run_studio_installation` and `run_studio_installer` are very similar and do not distinguish well with each other. Consider renaming `run_studio_installation` to `install_studio`

In the natural language, the following statement is very confusing: "to run studio installation, prepare studio installer and run studio installer". 
However you can say: "to install studio, you need to prepare installer and run it."

akihikodaki (approved):

PR #209, opened 2023-05-14
Updates for running Win11 SVVP

kostyanf14:


YanVugenfirer (approved):

lib/cli.rb
```diff
@@ -24,7 +24,7 @@ def initialize
 
     # class CommonOptions
     class CommonOptions
-      attr_accessor :debug, :config, :client_world_net
+      attr_accessor :debug, :config, :client_world_net, :fw_type
```
akihikodaki, line 27: I wonder if it's worth adding an option. Is it really more significant than other platform configurations like `machine_type`?

kostyanf14: Currently, we need this. To install Windows 11 PC should have TPM and SecureBoot enabled but for running SVVP tests SecureBoot should be disabled (except one test). Another solution is to create 2 platforms that use the same client images but different FWs. It looks strange that other platforms use the same image but redefining some states from CLI is ok (depending on requirements).

What solution is better from your point of view?

kostyanf14: One more thing. When SB is enabled, Windows prevents enabling test signing. This is ok for SVVP because all our drivers should be signed and HLK enables test signing on tests demand. 
During driver testing we install unsigned drivers, so test signing should be enabled before driver installation. Currently, HCK-CI does not have methods for this action and expects that test signing was enabled during the installation process.

akihikodaki: What about adding some code to enable SecureBoot only at the installation time? Doesn't that cover all the use case, or is it significantly complex than adding a CLI option?

kostyanf14: We only need to use proper OVMF to enable SecureBoot. So theoretically we can do this for all platforms with UEFI FW.

kostyanf14: But then we should check that during unsigned driver testing the testsigning is enabled

akihikodaki: It's not necessary to do this for *all* platforms. Instead we can add e.g., `installer_fw_type` field only for Windows 11, which can run the installer only with Secure Boot. That shouldn't break test signing for the other platforms.

akihikodaki:
Shouldn't `drivers` option be also defined as a platform configuration parameter instead of a CLI option?

kostyanf14:
> Shouldn't `drivers` option be also defined as a platform configuration parameter instead of a CLI option?

No. Currently, we support only `no driver` devices in this scenario that are enough for general installation (platform related). The next step is fixing specific tests. For example, use vioscsi as a boot device. We should install a driver during Windows image installation and this case is not a platform-specific.

akihikodaki:
What about always installing e.g., `vioscsi` then?

kostyanf14:
> What about always installing e.g., `vioscsi` then?

We can't. If we want to test vioscsi driver we should reinstall the driver for test one but Windows always use the proper signed driver (if it is present). So if we install a signed driver at the installation step, we can't test an unsigned driver.

akihikodaki:
Alright, that makes sense.

lib/cli.rb
```diff
@@ -176,7 +176,7 @@ def gthb_context_suffix_option(parser)
 
     # class InstallOptions
     class InstallOptions
-      attr_accessor :platform, :force, :skip_client
+      attr_accessor :platform, :force, :skip_client, :drivers
```
akihikodaki, line 179: What about having an empty array as the default value for `drivers` by default? That saves some code later, and sounds more semantically correct for me.

akihikodaki (changes requested):

lib/cli.rb
kostyanf14: good idea. Fixed

YanVugenfirer (approved):

PR #211, opened 2023-05-16
Generate a human-readable html report during test run

kostyanf14:
`results.html`:
image


akihikodaki (changes requested):
I don't think it's worth to parse and generate the summary of systeminfo. The log is already quite a verbose and the increase due to printing the entire systeminfo should be negligible. If we need to reduce the verbosity of the logs, it should be probably done by reducing repetitive logs somewhere instead.

kostyanf14:
`systeminfo` contains a lot of redundant information for this log: nic information (multiline), update information (multiline), directories, locales, etc. This information can be useful for debugging, but not as a part of the summary log.

@YanVugenfirer is it ok to add full  `systeminfo` output to summary log?

akihikodaki:
I agree it's unnecessary to print those extra information and there is an opportunity to reduce the verbosity. The question is whether this opportunity should be exploited, and my answer is no. The reduction is too little, considering the entire amount of the logs.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,186 @@
...
+        k = k.chomp(DELIM)
+        case k
+        # network_cards currently are not parsed
+        when 'hotfixs', 'processors'
+          raw_output[k] = parse_hotfixs_or_processors(v)
+        when 'hyperv_requirements'
+          raw_output[k] = parse_hyperv_requirements(v)
+        when 'total_physical_memory', 'available_physical_memory', 'virtual_memory_max_size',
+             'virtual_memory_available', 'virtual_memory_in_use'
+          raw_output["#{k}_mb"] = UtilsParser.convert_to_int(v)
```
akihikodaki, line 102: The methods called here are so complex. I suggest to remove them and add later if necessary.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,186 @@
...
+    #
+    # @param data [String] Text data to parse
+    #
+    # @return [Array<Hash>] List of dictionaries. Processed structured data.
+    def parse(data)
+      lines = data.lines.map(&:chomp).reject(&:empty?)
+
+      # find the character position of the value in the k/v pair of the first line
+      # all subsequent lines of data use the same character position
+      start_value_pos = get_value_pos(lines[0], DELIM)
```
akihikodaki, line 48: `systeminfo` has an option to CSV so you should use it when parsing mechanically.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,186 @@
...
+    # @param key [String] Input value
+    # @return [String] The transformed key
+    def transform_key(key)
+      # lowercase and replace spaces with underscores
+      key = key.strip.downcase.gsub(' ', '_')
+
+      # remove invalid key characters
+      ";:!@\#$%^&*()-".each_char { |c| key = key.gsub(c, '') }
+      key
+    end
```
akihikodaki, line 184: I don't think this is necessary.

lib/engines/hcktest/tests.rb
```diff
@@ -349,7 +353,33 @@ def create_project_package
...
     end
 
+    def generate_summary_system_info(info)
+      "Guest system information\n  " \
+        "Host: #{info['host_name']}\n  " \
+        "OS: #{info['os_name']} #{info['os_version']}\n  " \
+        "System: #{info['system_manufacturer']} #{info['system_model']} #{info['system_type']}\n  " \
+        "CPU: #{info['processors'].join(' ')}\n  " \
+        "Memory: #{info['total_physical_memory_mb']}\n  " \
+        "BIOS: #{info['bios_version']}\n"
```
akihikodaki, line 363: Isn't BIOS information enough? I think the other information can be inferred from the configuration and is useful only when debugging.

akihikodaki (changes requested):

lib/auxiliary/sysinfo_parser.rb
kostyanf14: CSV format is not fully helpful. Fields hotfixes, nic, processors, etc are still is a multiline data and some of them should be parsed.

lib/auxiliary/sysinfo_parser.rb
kostyanf14: Currently we need to parse a processors at least.

lib/engines/hcktest/tests.rb
kostyanf14: You are partially right but we still need CPU, OS and BIOS. 

When CPU is host we can't get the real value from config and QE reported some bugs related to Intel only. So at least we need know is this Intel or AMD.
OS information contains Windows version and build we need this for  testing the vnext/preview Windows version.

lib/auxiliary/sysinfo_parser.rb
kostyanf14: Not sure about this. How to add tests in this case. Some keys contains '()' for example.

lib/auxiliary/sysinfo_parser.rb
akihikodaki: Yes, but I think it will still save several lines and interface stability.

lib/auxiliary/sysinfo_parser.rb
akihikodaki: That makes sense. Remove the others then.

lib/auxiliary/sysinfo_parser.rb
akihikodaki: Probably it's fine even if parentheses are present.

lib/engines/hcktest/tests.rb
akihikodaki: The reasoning for CPU makes sense.

I'm not sure about OS. If you are *testing* vnext/preview then you should be fully aware that you are doing something unusual event without this log.

You can use heredoc here by the way.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+            Integer(Float(str_val))
+          rescue ArgumentError, TypeError
+            nil
+          end
+        end
+      elsif value.is_a?(Numeric)
+        value.to_i
+      end
+    end
+  end
```
akihikodaki, line 33: This is no longer used.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+    # @return [Array<Hash>] List of dictionaries. Processed structured data.
+    def parse(data)
+      rows = CSV.parse(data, headers: true)
+      raw_data = rows.first.to_h
+
+      raw_output = {}
+      raw_data.each do |k, v|
+        raw_output[k] = case k
+                        when 'Hotfix(s)', 'Processor(s)'
+                          parse_hotfixs_or_processors(v)
```
akihikodaki, line 52: It seems "Network Card(s)" also use the same syntax so perhaps it's better to include it and name the method: `parse_array`.

lib/engines/hcktest/tests.rb
```diff
@@ -349,9 +378,41 @@ def create_project_package
...
+
+      client_sysinfo = parser.parse(@tools.get_machine_system_info(@client.name))
+      build_system_info(client_sysinfo)
+
+      return if @support.nil?
+
+      support_sysinfo = parser.parse(@tools.get_machine_system_info(@support.name))
+      build_system_info(support_sysinfo)
+    end
+
```
akihikodaki, line 407: The store variable is named `guest_systems_info` but the method name says `clients_system_info`. Keep them consistent.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+    def parse_hotfixs_or_processors(data)
+      arr_output = []
+      data.split(',').each_with_index do |l, i|
+        # skip line that says how many are installed
+        next if i.zero?
+
+        # we have to make sure this is a complete line
+        # as data could cutoff. Make sure the delimiter
+        # exists. Otherwise, skip it.
+        next unless l.include?(DELIM)
```
akihikodaki, line 82: Can that cut off really happen, especially when it's told to output a machine-readable CSV?

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+      rows = CSV.parse(data, headers: true)
+      raw_data = rows.first.to_h
+
+      raw_output = {}
+      raw_data.each do |k, v|
+        raw_output[k] = case k
+                        when 'Hotfix(s)', 'Processor(s)'
+                          parse_hotfixs_or_processors(v)
+                        else
+                          v.strip
```
akihikodaki, line 54: Is this `strip` really necessary?

akihikodaki (changes requested):

lib/auxiliary/sysinfo_parser.rb
kostyanf14: Similar but not the same.
Processors: `4 Processor(s) Installed.,[01]: Intel64 Family 6 Model 42 Stepping 7 GenuineIntel ~3392 Mhz,[02]: ....`
Network: `2 NIC(s) Installed.,[01]: vmxnet3 Ethernet Adapter,      Connection Name: Local Area Connection 3,      DHCP Enabled:    No,      IP address(es),      [01]: 10.20.1.10,   ...`

Network Card(s) is a complete structure with a lot of `,` separators for internal object parts.

lib/auxiliary/sysinfo_parser.rb
kostyanf14: Yes. From one PC `"Microsoft Windows Server 2008 R2 Standard "`

lib/auxiliary/sysinfo_parser.rb
kostyanf14: droped

lib/engines/hcktest/tests.rb
kostyanf14: fixed

lib/auxiliary/sysinfo_parser.rb
kostyanf14: droped

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,61 @@
...
+      data.split(',').each_with_index do |l, i|
+        # skip line that says how many are installed
+        next if i.zero?
+
+        _, v = l.split(DELIM)
+        # discard the number sequence
+        arr_output << v.strip
+      end
+
+      arr_output
```
akihikodaki, line 58: I suggest to write the literal for `:` instead of referring to a constant as you do for `','`. It makes less sense to define a class-wide constant for something only used in this method, and the name is also confusing, considering `,` is also a _delimiter_.

Destructing assignment is used to extract the value, but I think the index syntax is simpler here as you only need an element.

Ruby provides various methods for array manipulation so use them to make it simpler.

You should specify the limit parameter when splitting the index and value so that it won't misbehave when the value includes `:`.

Wrapping up all of the suggestions:

```suggestion
      arr_output = data.split(',')

      # skip line that says how many are installed
      arr_output.shift

      # discard the number sequence
      arr_output.map! { _1.split(':')[1].strip }
```

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,61 @@
...
+      raw_data.each do |k, v|
+        raw_output[k] = case k
+                        when 'Hotfix(s)', 'Processor(s)'
+                          parse_hotfixs_or_processors(v)
+                        else
+                          v.strip
+                        end
+      end
+
+      raw_output
```
akihikodaki, line 32: To derive a `Hash` from `CSV::Row`:

```suggestion
      CSV.parse(data, headers: true).first.each.to_h do |k, v|
        [
          k,
          case k
          when 'Hotfix(s)', 'Processor(s)'
            parse_hotfixs_or_processors v
          else
            v.strip
          end
        ]
      end
```

akihikodaki (changes requested):

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,52 @@
...
+module AutoHCK
+  # SysInfoParser class
+  class SysInfoParser
+    DELIM = ':' # k/v delimiter
+
+    # Main text parsing function
+    #
+    # @param data [String] Text data to parse
+    #
+    # @return [Array<Hash>] List of dictionaries. Processed structured data.
```
akihikodaki, line 17: It should be just a `Hash`, not an `Array` of `Hash`es.

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'csv'
+require 'active_support'
+require 'active_support/core_ext/object/deep_dup'
+
+# AutoHCK module
+module AutoHCK
+  # SysInfoParser class
+  class SysInfoParser
+    DELIM = ':' # k/v delimiter
+
```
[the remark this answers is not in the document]
kostyanf14: removed

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,52 @@
...
+    # Main text parsing function
+    #
+    # @param data [String] Text data to parse
+    #
+    # @return [Hash] List of dictionaries. Processed structured data.
```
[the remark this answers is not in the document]
kostyanf14: fixed

lib/auxiliary/sysinfo_parser.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'csv'
+require 'active_support'
+require 'active_support/core_ext/object/deep_dup'
```
akihikodaki, line 5: Sorry for repeatedly adding more comments, but I found more unused code here.

akihikodaki (changes requested):

lib/auxiliary/sysinfo_parser.rb
kostyanf14: we need a review to make the code cleaner, so not a problem

kostyanf14: removed

YanVugenfirer (approved):

PR #216, opened 2023-06-08
Updates for running Win11 Insider SVVP

kostyanf14:


akihikodaki (changes requested):
- The names `hlk_playlist` and `--playlist` do not distinguish each other. `--playlist` also specifies something related to HLK. The difference is that `--playlist` only contains the names.
- Giving the `--playlist` command line option makes AutoHCK ignore the XML playlist, but the `playlist` target configuration does not have the same effect. Is there justification for this?
- `pre_start_once_commands` are added to the "manual" shell script, but that would prevent from repetitively running it. What about just running those commands in AutoHCK even for manual execution?

akihikodaki:
> > * The names `hlk_playlist` and `--playlist` do not distinguish each other. `--playlist` also specifies something related to HLK. The difference is that `--playlist` only contains the names.
> 
> I do not agree with you. `hlk_playlist` is for Microsoft official HLK playlists, `playlist` is a HCK-CI/AutoHCK playlist. Do you have any naming suggestions? @YanVugenfirer What do you think?

My understanding is that the studio machine has HLK tests, and `hlk_playlist` filters them with various conditions described in XML and `playlist` filters them only with names. This way, `playlist` is as related to HLK as `hlk_playlist` is.

> 
> > ```
> > * Giving the `--playlist` command line option makes AutoHCK ignore the XML playlist, but the `playlist` target configuration does not have the same effect. Is there justification for this?
> > ```
> 
> Sorry, I am not sure that understand you correctly. Do you mean that the `playlist` in the `driver.json` and the `--playlist` option have a different behavior (with option AutoHCK ignore XML)? If yes, this is my mistake.

I was wrong, I read the following lines incorrectly, but these lines still do not make sense.

```
      return if @project.options.test.playlist.nil?

      @kit_playlist = ms_playlist(true)
```

https://github.com/HCK-CI/AutoHCK/pull/216/files#diff-7e61964988af0c20dce2999273920ae120d376eb50c92f8d7314ca83813d431dR27-R29

`@kit_playlist` will be `nil` if `playlist` option is not specified, but it does not seem right.

> 
> > ```
> > * `pre_start_once_commands` are added to the "manual" shell script, but that would prevent from repetitively running it. What about just running those commands in AutoHCK even for manual execution?
> > ```
> 
> I am not sure about this. One of my personal usage behavior is:
> 
>     * generate files
> 
>     * start scripts and debug driver / HLK
> 
>     * go sleep and shutdown PC
> 
>     * start PC
> 
>     * start scripts and continue debugging in the same env
> 
> 
> in this case, I got a problem.
> 
> @YanVugenfirer @ybendito What do you think about this? And how do you see the usage?

My usage is also similar. The manual execution allows to use the same environment repetitively. In such a case, `pre_start_once_commands` should not be executed when running scripts but when generating files.

kostyanf14:
>Giving the --playlist command line option makes AutoHCK ignore the XML playlist, but the playlist target configuration does not have the same effect. Is there justification for this?

Fixed

YanVugenfirer:
Hi,
Playlists naming is confusing in retrospective.
Maybe emphasise the difference by naming the something like ms_playlist (can keep hlk_playlist too) and custom_playlist (or user_playlist).

lib/engines/hcktest/playlist.rb
```diff
@@ -17,15 +17,21 @@ def initialize(client, project, target, tools, kit)
       @tools = tools
       @logger = project.logger
       @kit = kit
-      @ms_playlist = ms_playlist(true)
+      @kit_playlist = nil
```
akihikodaki, line 20: This is no longer needed.

```suggestion
```

lib/engines/hcktest/playlist.rb
```diff
@@ -60,7 +77,14 @@ def sort_by_duration
     end
 
     def custom_playlist(log)
-      playlist = @project.engine.target['playlist']
+      custom_playlist_file = @project.options.test.playlist
+
+      playlist = if custom_playlist_file.nil?
+                   @project.engine.target['playlist']
+                 else
+                   File.readlines(custom_playlist_file).map(&:chomp)
```
akihikodaki, line 85: `File.readlines` has `chomp` option so use it.

lib/engines/hcktest/playlist.rb
```diff
@@ -69,19 +93,20 @@ def custom_playlist(log)
...
+
+      ignore_list = if custom_ignore_list_file.nil?
+                      @project.engine.target['ignore_list']
+                    else
+                      File.readlines(custom_ignore_list_file).map(&:chomp)
+                    end
+
+      return unless ignore_list
+
+      @ignored_list = @tests.filter { ignore_list.include?(_1['name']) }
```
akihikodaki, line 106: I prefer `reject` rather than `filter` as the old code did. `filter` is somewhat unclear if it's selecting or rejecting.

akihikodaki (changes requested):

lib/engines/hcktest/playlist.rb
kostyanf14: In the case of old code with `reject` after adding a condition for `File.readlines` we have a big `rubocop metrics/abcsize`. That's why I rewrite the mechanism. 

`@ignored_list = @tests.filter { ignore_list.include?(_1['name'])` - select ignored test 
`@tests -= @ignored_list` - reject ignored tests from list

kostyanf14: simplified old source and back logic to use `reject`

lib/engines/hcktest/playlist.rb
kostyanf14: fixed

lib/engines/hcktest/playlist.rb
kostyanf14: fixed

kostyanf14:
> Hi, Playlists naming is confusing in retrospective. Maybe emphasise the difference by naming the something like ms_playlist (can keep hlk_playlist too) and custom_playlist (or user_playlist).

@YanVugenfirer 

Updated options: `--ms-playlist` and `--user-playlist`

kostyanf14:
@akihikodaki 

Am I right that we have only one open question related to `pre_start_once_commands`?

lib/engines/hcktest/playlist.rb
akihikodaki: I didn't meant to use `reject!` but to use `reject`, which is an alias of `filter`. Either of `reject` and `reject!` is fine in my opinion.

akihikodaki:
I'm still not convinced about the naming of `--ms-playlist` and `--user-playlist`. The naming has three problems:

1. `--ms-playlist` can point to a XML the user created instead of one provided by Microsoft.
2. The word `playlist` is overloaded.

It's indeed likely that `--ms-playlist` points to something Microsoft provided and `--user-playlist` points to something the user came up, but AutoHCK provides filtering functionality for both of the options; AutoHCK does not care if `--ms-playlist` points to a XML provided by Microsoft or a XML user created. The internal method name `ms_playlist` made sense before this change as it always points to XMLs Microsoft provides, but now it's specified by the user so it may not be from Microsoft.

Problem 2 was not a big problem before this change as `--ms-playlist` and `--user-playlist` were specified in different files: platform configuration file and driver configuration file. However, AutoHCK now allows to specify them in the command line altogether so differentiating them is important now. In the first place, "playlist" is a well-defined name in HLK to denote a XML format so it's better to avoid to use the word for `--user-playlist`.

`--user-playlist` is more similar to `--ignore-list` in terms that it only specifies test names in a plain text so it's better to align with it instead so it can be `--accept-list`, for example. The word "list" still does not tell it apart from the word "playlist"; it's better to describe that the lists to specify with the options are the names of tests. For example, they can be named `--ignore-names` and `--accept-names`.

Regarding `pre_start_once_commands` is there any concern about running the commands in AutoHCK when generating files for manual execution?

kostyanf14:
Conclusions after discussion: 
- use option `playlist` for MS/XML playlist
- use option `select-test-names` for plain text playlist
- use option `reject-test-names` for plain text ignore-list
- rename all `ignore-list` related code to be consistent with the option
- use `config_commands` instead of `pre_start_once_commands` 
- generate separate `config_manual.sh` for `config_commands` in manual mode

lib/engines/hcktest/playlist.rb
```diff
@@ -52,36 +57,61 @@ def ms_playlist(log)
...
-      if ignore_list
-        @tests.reject! do |test|
-          if ignore_list.include?(test['name'])
-            @ignored_list << test
-            true
-          else
-            false
-          end
+    def custom_reject_test_names(log)
+      custom_reject_test_names_file = @project.options.test.reject_test_names
```
akihikodaki, line 95: The word `custom` is overloaded; it intended to denote it's from user, here but the method name also includes the word to describe it's not from XML. Instead, use the word `user` as `custom_select_test_names` does.

```suggestion
      user_reject_test_names_file = @project.options.test.reject_test_names
```

lib/engines/hcktest/playlist.rb
```diff
@@ -52,36 +57,61 @@ def ms_playlist(log)
...
         @tests[i]['duration'] = time_to_seconds(x['estimatedruntime'])
       end
       @tests.sort_by! { |test| test['duration'] }
     end
 
-    def custom_playlist(log)
-      playlist = @project.engine.target['playlist']
-      return unless playlist
+    def custom_select_test_names(log)
+      user_test_names_file = @project.options.test.select_test_names
```
akihikodaki, line 79: Include the word `select` for consistency.

```suggestion
      user_select_test_names_file = @project.options.test.select_test_names
```

akihikodaki (changes requested):

akihikodaki (approved):

YanVugenfirer (approved):

PR #220, opened 2023-06-28
Uploaders rework

kostyanf14:


lib/resultuploaders/dropbox.rb
```diff
@@ -5,33 +5,37 @@
...
-
   # dropbox class
   class Dropbox
+    TOKEN_JSON = 'lib/resultuploaders/dropbox.token.json'
+    CONFIG_JSON = 'lib/resultuploaders/dropbox.json'
+
     include Helper
 
-    attr_reader :url
+    attr_reader :url, :html_url
```
akihikodaki, line 15: If you want a method that returns `nil`, just write: `def html_url; end`

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,48 @@ class ResultUploader
     # UploaderFactory
     #
     class UploaderFactory
-      UPLOADERS = {
-        dropbox: Dropbox
-      }.freeze
+      def self.initialize(logger)
+        @uploades = {}
```
akihikodaki, line 15: Not `uploaders`?

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,48 @@ class ResultUploader
...
+
+        Dir['./lib/resultuploaders/*.rb'].sort.each do |file|
```
akihikodaki, line 17: Strictly I don't think it's assured that the current working directory is at the project root.

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,48 @@ class ResultUploader
     # UploaderFactory
     #
     class UploaderFactory
-      UPLOADERS = {
-        dropbox: Dropbox
-      }.freeze
+      def self.initialize(logger)
```
akihikodaki, line 14: Why did you put this in a class method? can't you just run the code in the body of UploaderFactory?

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,48 @@ class ResultUploader
...
+        @uploades = {}
+
+        Dir['./lib/resultuploaders/*.rb'].sort.each do |file|
+          uploader_name = File.basename(file, '.rb')
+          # Ignore current file
+          next if uploader_name == File.basename(__FILE__, '.rb')
```
akihikodaki, line 20: What about creating a new directory `driver` and putting each uploader implementations into that directory? This will simplify the logic and clarifies this file and the other "driver" files have different roles.

akihikodaki (changes requested):

lib/resultuploaders/result_uploader.rb
akihikodaki: Oh, I missed the logger usage. In my opinion:
- Let `try_load_uploader` fail when a class is not found. Such a case is almost certainly a bug.
- Let the default exception handler handle load errors. It should tell the context enough to debug in case of failure.

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,49 @@ class ResultUploader
     # UploaderFactory
     #
     class UploaderFactory
-      UPLOADERS = {
-        dropbox: Dropbox
-      }.freeze
+      def self.initialize(logger)
+        @uploaders = {}
+
+        Dir["#{__dir__}/*"].each do |dir|
```
akihikodaki, line 17: This is problematic if `__dir__` contains `*` though it's very unlikely. Use `Dir#each_child`.

akihikodaki (changes requested):

lib/resultuploaders/result_uploader.rb
kostyanf14: You are right. The default exception handler output is good for debugging.

lib/resultuploaders/result_uploader.rb
kostyanf14: Fixed

lib/resultuploaders/dropbox/dropbox.rb
```diff
@@ -20,20 +20,25 @@ def initialize(project)
...
 
+    def html_url; end
+
+    def load_config
+      @config = Json.read_json(CONFIG_JSON, @logger)
+      @action_retries = @config['action_retries']
+      @action_retry_sleep = @config['action_retry_sleep']
+
+      @client_id = ENV.fetch('AUTOHCK_DROPBOX_CLIENT_ID')
+      @client_secret = ENV.fetch('AUTOHCK_DROPBOX_CLIENT_SECRET')
```
akihikodaki, line 39: `@client_id` and `@client_secret` are somewhat out of place; they are not from `CONFIG_JSON`. You may put them after `load_config`, or just abandon this change to introduce `load_config` method; I suspect this method is introduced to suppress the complexity warning caused by `@html_url` assignment, but `html_url` implementation is now stateless so it's no longer necessary.

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,24 +11,38 @@ class ResultUploader
     # UploaderFactory
     #
     class UploaderFactory
-      UPLOADERS = {
-        dropbox: Dropbox
-      }.freeze
+      @uploaders = {}
```
akihikodaki, line 14: Now I see no reason for this conversion from constant to variable.

lib/resultuploaders/result_uploader.rb
```diff
@@ -73,9 +88,11 @@ def update_file_content(content, r_name)
     end
 
     def url
-      return nil unless @connected_uploaders.key?(:dropbox)
+      @connected_uploaders.values.filter_map(&:url).first
```
akihikodaki, line 91: What about just using `find`?

akihikodaki (changes requested):

lib/resultuploaders/result_uploader.rb
kostyanf14: Sorry, what do you mean? 
Drop `@uploaders` and rewrite the `create` function? 

```ruby
      def self.create(type, project)
        AutoHCK.const_get(type.camelize).new(project)
      end
 ```

File: lib/resultuploaders/result_uploader.rb (hunk 9 above)
Comment (kostyanf14) [reply lib/resultuploaders/result_uploader.rb:91]: Variants `.map(&:url).find { !_1.nil? }` and `.find { !_1.url.nil? }&.url` looks more complicated to me.

File: lib/resultuploaders/result_uploader.rb (hunk 8 above)
Comment (akihikodaki) [reply lib/resultuploaders/result_uploader.rb:14]: You can write:

```suggestion
      UPLOADERS = {}
```

lib/resultuploaders/result_uploader.rb
akihikodaki: Ah, makes sense.

lib/resultuploaders/dropbox/dropbox.rb
kostyanf14: 1. You are right `load_config` is not required for now.
2. `@client_id` and `@client_secret` are also configurations of the DropBox client, but these are secret values, so they pleased into ENV instead of `config.json`

lib/resultuploaders/result_uploader.rb
kostyanf14: Yep

lib/resultuploaders/dropbox/dropbox.rb
akihikodaki: > 1. You are right `load_config` is not required for now.
> 
>     2. `@client_id` and `@client_secret` are also configurations of the DropBox client, but these are secret values, so they pleased into ENV instead of `config.json`

But the data loaded from a file is what is named `@config`. Perhaps `@config` is what should be renamed.

lib/resultuploaders/result_uploader.rb
kostyanf14: In this case of the UPLOADERS constant, we got the RuboCop offense `Freeze mutable objects assigned to constants.Style/MutableConstant(RuboCop)`. We can't freeze this hash because it will be updated.

akihikodaki: You can (and perhaps you should) freeze it after the hash is filled.

kostyanf14: But RuboCop asked to freeze during objects assigned.

kostyanf14: No, because we have return for files. to_h fail in this case, or you suggest to return [nil,nil] in this case?

akihikodaki: You can first have an `Enumerator` for paths under `__dir__`, and then call `select` to select directories.

kostyanf14: Fixed

lib/resultuploaders/dropbox/dropbox.rb
kostyanf14: Fixed

lib/resultuploaders/result_uploader.rb
```diff
@@ -10,9 +11,20 @@ class ResultUploader
...
+        file = "#{full_path}/#{uploader_name}.rb"
+        require file
+
+        # Convert 'result_uploader' (file name) -> 'ResultUploader' (class name)
+        class_name = uploader_name.camelize
+        # Convert 'ResultUploader' (class name) -> ResultUploader (declared constant)
+        # or
+        # Convert 'ResultUploader' (class name) -> AutoHCK::ResultUploader (declared constant)
+        [uploader_name, AutoHCK.const_get(class_name)]
+      end.freeze
```
akihikodaki, line 27: Chaining `filter_map` and `to_h` will be simpler.

akihikodaki (changes requested):

lib/resultuploaders/result_uploader.rb
akihikodaki:
```suggestion
      UPLOADERS = Dir.each_child(__dir__)
                     .filter_map do |uploader_name|
        full_path = "#{__dir__}/#{uploader_name}"        
        next unless File.directory? full_path

        file = "#{full_path}/#{uploader_name}.rb"
        require file

        # Convert 'result_uploader' (file name) -> 'ResultUploader' (class name)
        class_name = uploader_name.camelize
        # Convert 'ResultUploader' (class name) -> ResultUploader (declared constant)
        # or
        # Convert 'ResultUploader' (class name) -> AutoHCK::ResultUploader (declared constant)
        [uploader_name, AutoHCK.const_get(class_name)]
      end.to_h.freeze
```

akihikodaki (approved):

YanVugenfirer (approved):

PR #227, opened 2023-07-12
Introduce AutoHCK::ResourceScope

akihikodaki:
I have already done several changes to improve the abortion logic, but I still see an error occurs during abortion. I also heard AutoHCK sometimes hangs after the tests finish, which imply some resources associated threads are left unreleased.

To eliminate this kind of errors completely, I propose `AutoHCK::ResourceScope` as a single true resource management mechanism.

This pull request makes *every* code dealing with unmanaged resources use the new mechanism so it's huge. Take time if it's correctly used everywhere before merging.

### DIFF
--- a/bin/auto_hck
+++ b/bin/auto_hck
@@ -7,6 +7,7 @@ require './lib/config/sentry'
 
 begin
   require 'filelock'
+  require './lib/auxiliary/resource_scope'
   require './lib/cli'
   require './lib/project'
   require './lib/trap'
@@ -25,8 +26,8 @@ begin
     Thread.abort_on_exception = true
     Thread.report_on_exception = false
 
-    @project = Project.new(cli)
-    begin
+    ResourceScope.open do |scope|
+      @project = Project.new(scope, cli)
       Trap.project = @project
       @project.run if @project.prepare
     rescue StandardError => e
@@ -37,9 +38,9 @@ begin
       @project.log_exception(e, 'fatal')
       @project.handle_error
       exit(1)
-    ensure
-      @project.close
     end
+  rescue AutoHCKInterrupt
+    exit false
   end
 rescue StandardError => e
   Sentry.capture_exception(e)
--- a/lib/engines/hcktest/hcktest.rb
+++ b/lib/engines/hcktest/hcktest.rb
@@ -144,10 +144,14 @@ module AutoHCK
       Json.read_json(platform_json, @logger)
     end
 
-    def initialize_clients
+    def run_studio(run_opts = {})
+      @studio = @project.setup_manager.run_hck_studio(run_opts)
+    end
+
+    def run_clients(run_opts = {})
       @clients = {}
       @platform['clients'].each do |_name, client|
-        @clients[client['name']] = @project.setup_manager.create_client(client['name'])
+        @clients[client['name']] = @project.setup_manager.run_hck_client(client['name'], run_opts)
 
         break if @project.options.test.svvp
         break unless @drivers.any? { |d| d['support'] }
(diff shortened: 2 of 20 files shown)


akihikodaki:
Also: I'm still testing and have not confirmed its well-behaving.

lib/setupmanagers/hckclient.rb
```diff
@@ -17,17 +17,14 @@ class HCKClient
...
       @logger = project.logger
       @studio = studio
       @name = name
       @kit = setup_manager.kit
       @setup_manager = setup_manager
       @pool = 'Default Pool'
-    end
-
-    def run(run_opts = nil)
       @setup_manager.run_client(@name, run_opts)
```
kostyanf14: As these methods will be more complicated, I suggest adding several debug logs to know what we exactly do now.
And in hckstudio.

lib/setupmanagers/hckclient.rb
```diff
@@ -201,10 +197,9 @@ def reset_to_ready_state
       set_machine_ready
     end
 
-    def abort
+    def close
       @logger.info("Aborting HLKClient #{@name}")
```
kostyanf14, line 201: "Closing" instead of "Aborting" ?

lib/engines/hcktest/hcktest.rb
```diff
@@ -183,31 +184,25 @@ def configure_setup_and_synchronize
...
-      configure_setup_and_synchronize
+        configure_setup_and_synchronize
+        tmp_scope.move_to scope
+      end
     rescue AutoHCKError => e
-      synchronize_clients(exit: true)
       @project.logger.warn("Running and configuring setup failed: (#{e.class}) #{e.message}")
       raise e unless (retries += 1) < AUTOHCK_RETRIES
 
-      clean_last_run_machines
```
kostyanf14, line 211: Will we recreate snapshots when configuring setup failed? Sorry, this is not clear to me.

kostyanf14 (commented):
Very good job. Have several questions for now. Will be back after testing.

lib/engines/hcktest/hcktest.rb
akihikodaki: Yes, when a setup fails, it will first close the current resource scope, which makes snapshots deleted. Then, if the retry counter is not saturated, it will rerun from the start, creating new snapshots.

lib/setupmanagers/hckclient.rb
akihikodaki: Done with commit 9bbd4ae218bf81aeae32584f1b00da9ace886ffe.

lib/setupmanagers/hckclient.rb
akihikodaki: Done with commit 0ff4aa59654e0ba1b04be37fa23276b6a43f03e9, but I used word "exiting" instead; it's a bit awkward to say closing something not a file. I just named the method "close" to match the interface; there is no such restriction for a log message.

akihikodaki:
It happened to hang with old Timeout so I fixed it with commit 1b276f349ae1c1df4c1d83a94f115e1ac0813045.
It did not hang with a recent Timeout because of a regression. For details of the regression, see: https://github.com/ruby/timeout/issues/41

akihikodaki:
Commit 4b59a081c1f9b34be5f84e9ffc955c9e1dc0e362 removed destructive methods from `AutoHCK::ResourceScope` so that you can pass its instances around without worrying that resources already allocated may be destructed somewhere you are unaware of.

bin/auto_hck
```diff
@@ -7,6 +7,7 @@ require './lib/config/sentry'
```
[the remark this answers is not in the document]
akihikodaki: Done with commit ef5507463b760957b3ac2c84debc6eb6f73d7ba0.

kostyanf14 (approved):

YanVugenfirer (approved):

PR #228, opened 2023-07-13
HTML report enhancement

kostyanf14:


lib/cli.rb
```diff
@@ -197,6 +199,23 @@ def reject_test_names_option(parser)
           @reject_test_names = reject_test_names
         end
       end
+
+      def reject_report_section_option(parser)
+        parser.on('--reject-report-sections <reject_report_section>', Array,
+                  'List of section to reject from HTML results\n (use ' \
```
akihikodaki, line 205: This `\n` will not be a newline as it's single-quoted.

I also suggest to end the first literal with `\n`; it should be easier to read in this way. Perhaps you may use a heredoc, but I worry the syntax will be too complicated due to the block.

lib/cli.rb
```diff
@@ -85,7 +86,7 @@ def version_option(parser)
     class TestOptions
       attr_accessor :platform, :drivers, :driver_path, :commit, :diff_file, :svvp, :manual,
                     :gthb_context_prefix, :gthb_context_suffix, :playlist, :select_test_names,
-                    :reject_test_names
+                    :reject_test_names, :reject_report_section
```
akihikodaki, line 89: The option is named `--reject-report-sections` and its value is an `Array`, but this attribute is not plural. It is confusing.

akihikodaki (changes requested):
Why will you need `--reject-report-sections`?

akihikodaki (approved):

PR #230, opened 2023-07-18
Add support to use vioscsi/vioblk as a boot device

kostyanf14:


lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -0,0 +1,116 @@
...
+        @logger.info("Creating CL#{@client_id} snapshot file")
+        run_cmd("#{@qemu_img_bin} create -f #{IMAGE_FORMAT} -F #{IMAGE_FORMAT} -b #{@base_image_path} #{snapshot_path}")
+      end
+
+      def delete_boot_snapshot
+        FileUtils.rm_f(snapshot_path)
+      end
+
+      def boot_device_command(device_name, run_opts, qemu_replacement_list = {})
+        type = __method__.to_s.split('_').first
```
akihikodaki, line 89: This is excessively complex. I understand it is here to remove the duplication of the word `boot`, but it introduces a duplication of `_` instead and moreover introduces some reflective programming construct and string manipulation. Overall it's not reducing but adding complexity.

I know it's from `NetworkManager` but I don't want another copies of this pattern. This variable is only used for a debug log in this class so I suggest to simply remove it.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -0,0 +1,116 @@
...
+        filename = File.basename(@base_image_path, '.*')
+        "#{@workspace_path}/#{filename}-snapshot.#{IMAGE_FORMAT}"
+      end
+
+      def check_image_exist
+        File.exist?(@base_image_path)
+      end
+
+      def create_image(path, size_gb)
+        run_cmd("#{@qemu_img_bin} create -f #{IMAGE_FORMAT} #{path} #{size_gb}G")
```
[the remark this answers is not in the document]
kostyanf14: I can, but I don't want to. During LOGO tests, HLK writes the whole disk so that host space decreases quickly. We use 30G because it is enough for all tests according to requirements.

lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -0,0 +1,112 @@
...
+          FileUtils.cp(@fs_test_image, @test_image_path)
+        else
+          @logger.info("Creating CL#{@client_id} test image")
+          create_image(@test_image_path, 30)
+        end
+      end
+
+      def create_boot_snapshot
+        @logger.info("Creating CL#{@client_id} snapshot file")
+        run_cmd("#{@qemu_img_bin} create -f #{IMAGE_FORMAT} -F #{IMAGE_FORMAT} " \
```
akihikodaki, line 80: I suggest using string-array literal here. It removes the need to escape the newline and avoids the shell.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -0,0 +1,113 @@
...
+          @logger.info("Creating CL#{@client_id} test image")
+          create_image(@test_image_path, 30)
+        end
+      end
+
+      def create_boot_snapshot
+        @logger.info("Creating CL#{@client_id} snapshot file")
+        cmd = [@qemu_img_bin, 'create -f', IMAGE_FORMAT, '-F', IMAGE_FORMAT,
+               '-b', @boot_image_path, boot_snapshot_path]
+        run_cmd(cmd.join(' '))
```
akihikodaki, line 82: You can pass the array as is.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/storage_manager.rb
kostyanf14: Why? I don't see any usage of `spawn` with an array.

akihikodaki: I was wrong, cmd_run cannot accept an array as a parameter. Please revert to the interpolated string literal.

kostyanf14: Is the current solution with an array worse?

akihikodaki: The old interpolated string literal looks more concise to me; it had fewer characters and no additional call for `Array#join`.

lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -0,0 +1,112 @@
...
+        @test_image_path = Pathname.new(@workspace_path).join("client#{@client_id}_test_image.#{IMAGE_FORMAT}")
+      end
+
+      def read_device(device)
+        @logger.info("Loading device: #{device}")
+        device_json = "#{QemuMachine::DEVICES_JSON_DIR}/#{device}.json"
+        unless File.exist?(device_json)
+          @logger.fatal("#{device} does not exist")
+          raise(InvalidConfigFile, "#{device} does not exist")
+        end
```
akihikodaki, line 36: You should check if `Json.read_json` raises the "file not found" error.

Generally, it's a bad practice to check a file's property before opening the file as it can change between the check and file opening. We do not really care such a race condition in this particular case, but it's better to always stick to the best practice.

I also wonder if this error is really necessary. Perhaps it may be fine to just let `Json.read_json` raise an error.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -27,7 +26,7 @@ def initialize(id, client_id, machine, logger)
 
       def read_device(device)
         @logger.info("Loading device: #{device}")
-        device_json = "#{DEVICES_JSON_DIR}/#{device}.json"
+        device_json = "#{QemuMachine::DEVICES_JSON_DIR}/#{device}.json"
```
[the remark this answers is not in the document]
kostyanf14: This is just simply removing duplicates. Any reason to drop this commit?

lib/setupmanagers/qemuhck/storage_manager.rb
kostyanf14: Fixed

YanVugenfirer (approved):

lib/setupmanagers/qemuhck/network_manager.rb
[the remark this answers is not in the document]
akihikodaki: You should be just able to refer to `AutoHCK::QemuMachine::DEVICES_JSON_DIR` as `DEVICES_JSON_DIR` here.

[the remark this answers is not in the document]
kostyanf14: You are right.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -303,13 +316,10 @@ def process_devices
         load_device_info(device_info)
       end
 
-      unless device_infos.any? { |d| d['type'] == 'vga' }
-        vga_info = read_device(@config['platforms_defaults']['vga_device'])
-        device_infos << vga_info
-        load_device_info(vga_info)
-      end
+      device_infos = add_missing_default_devices(device_infos)
```
akihikodaki, line 319: I prefer:
```suggestion
      add_missing_default_devices(device_infos)
```

Having an assignment for `device_infos` gives a false idea that the returned `Hash` is different from the original `Hash`, but they are actually the same.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: Fixed

PR #231, opened 2023-07-24
Fix error when reject_report_sections is nil

kostyanf14:


lib/engines/hcktest/tests.rb
```diff
@@ -223,7 +223,7 @@ def report_data
         'system_info' => {
           'guest' => @clients_system_info
         },
-        'sections' => RESULTS_REPORT_SECTIONS - @project.options.test.reject_report_sections
+        'sections' => RESULTS_REPORT_SECTIONS - @project.options.test.reject_report_sections.to_a
```
akihikodaki, line 226: Let's assign `[]` to @project.options.test.reject_report_sections just as we did for `AutoHCK::CLI::InstallOptions#drivers`.

akihikodaki (changes requested):

akihikodaki (approved):

YanVugenfirer (approved):

PR #249, opened 2023-09-14
Run pre start command before each QEMU run

kostyanf14:
When QEMU exits, device daemons exit too, so we should run it again before start QEMU.

### DIFF
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -101,6 +101,7 @@ module AutoHCK
                 Process.kill 'KILL', qemu.pid
                 qemu.wait_no_fail
               end
+              @machine.run_post_stop_commands
             end
 
             break unless @keep_alive
@@ -153,7 +154,6 @@ module AutoHCK
         return if @run_opts[:dump_only]
 
         vm_abort
-        @machine.run_post_stop_commands
         @machine.delete_snapshot if @delete_snapshot
       end
     end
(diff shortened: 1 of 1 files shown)


akihikodaki (changes requested):
This doesn't look correct. In the case of `ivshmem-doorbell`, the ivshmem server terminates during the execution of `post_stop_commands`, which only happens after `@qemu_thread` terminates. Repetitively running `pre_start_commands` will result in multiple ivshmem server processes running.

What daemons automatically quit after QEMU exits?

kostyanf14:
>What daemons automatically quit after QEMU exits?

virtiofsd and swtpm

akihikodaki (changes requested):
Let's move `@machine.run_post_stop_commands` into `@qemu_thread` then.

akihikodaki (changes requested):
`ensure` clause must be used so that `post_stop_commands` will be always executed.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -86,10 +86,11 @@ def run_qemu
 
       def run_vm
         @machine.dump_config
-        @machine.run_pre_start_commands
 
         @qemu_thread = Thread.new do
           loop do
+            @machine.run_pre_start_commands
```
akihikodaki, line 92: `@machine.run_pre_start_commands` should be in `Thread.handle_interrupt(Object => :on_blocking)` block. There is a possibility that `@machine.run_pre_start_commands` did some work that must be discarded with `              @machine.run_post_stop_commands` before raising an exception.

akihikodaki (changes requested):

akihikodaki (approved):

YanVugenfirer (approved):

PR #259, opened 2023-10-03
Add MSDN url for each test in html report

kostyanf14:


YanVugenfirer (approved):

akihikodaki (changes requested):
`report_data` now modifies tests, which does not look like a good idea.
- For example, consider that you are adding a new feature that outputs the URLs in the future. You would naturally `pp playlist.test` to inspect the test data, find the `url` property, and use it. Later, you will no longer need an HTML output so remove the call to `report_data`. When you do so, it accidentally breaks the irrelevant URL output you have added earlier.
- This computes URLs twice in `print_tests` and `report_data`, but you don't really have to when you are storing URLs.

A possible alternative way is to store URLs in `AutoHCK::Playlist#list_tests`. This way, you can always expect tests to include URLs.

akihikodaki (approved):

PR #266, opened 2023-11-02
Fail test run when QEMU fails too quickly

kostyanf14:


YanVugenfirer (approved):

akihikodaki (changes requested):
Use `Process#success?`.
Also, can't we just always raise an error when the exit status is not 0?

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -99,7 +113,7 @@ def run_vm
               @machine.run_pre_start_commands
 
               qemu = run_qemu(scope)
-              qemu.wait_no_fail
+              @fails_quickly = check_fails_too_quickly(qemu.wait_no_fail&.exitstatus)
```
akihikodaki, line 116: Why don't you raise immediately after `qemu = nil`?
You can also `@fails_quickly` turn into a local variable by doing so. It is better to have fewer states (instance variables).

It is also better to use `.` instead of `&.`; `AutoHCK::CmdRun::wait_no_fail` returns `nil` if the process does not terminate, but such a situation should not happen since `Process::WNOHANG` is not passed to `AutoHCK::CmdRun::wait_no_fail`. If `AutoHCK::CmdRun::wait_no_fail` returns `nil` despite that, it only implies the program has a bug. Assuming the process either succeeded or failed is wrong in that case, and we should raise an error instead.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -89,6 +89,20 @@ def run_qemu(scope)
         qemu
       end
 
+      def check_fails_too_quickly(status)
+        if status&.zero?
+          @first_fail_time = nil
+          false
+        end
+
+        if @first_fail_time.nil?
```
akihikodaki, line 98:
```suggestion
        elsif @first_fail_time.nil?
```

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -89,6 +89,20 @@ def run_qemu(scope)
...
+          @first_fail_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          false
+        else
+          (Process.clock_gettime(Process::CLOCK_MONOTONIC) - @first_fail_time) <= 10
```
akihikodaki, line 102:
```suggestion
          Process.clock_gettime(Process::CLOCK_MONOTONIC) - @first_fail_time <= 10
```

akihikodaki (changes requested):
I added comments for a few things I missed in the last review.

akihikodaki (approved):

PR #267, opened 2023-11-02
Wait for virtiofsd socket exist before start QEMU

kostyanf14:


lib/setupmanagers/qemuhck/devices/vhost-user-fs-pci.json
```diff
@@ -12,7 +12,8 @@
   "pre_start_commands": [
     "mkdir -p @fs_daemon_share_path@",
     "rm -f @fs_daemon_socket@",
-    "@fs_daemon_bin@ --socket-path=@fs_daemon_socket@ -o source=@fs_daemon_share_path@ -o cache=always &"
+    "@fs_daemon_bin@ --socket-path=@fs_daemon_socket@ -o source=@fs_daemon_share_path@ --cache always &",
+    "timeout 30 bash -c 'while [ ! -S \"@fs_daemon_socket@\" ]; do echo \"Socket does not exist, waiting...\"; sleep 1; done'"
```
akihikodaki, line 16: - Use `until`.
- It may be better to apply timeout for the entire initialization sequence on the Ruby side.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -548,16 +548,20 @@ def run_config_commands
     end
 
     def run_pre_start_commands
-      @pre_start_commands.each do |dirty_cmd|
-        cmd = replace_string_recursive(dirty_cmd, full_replacement_list)
-        run_cmd(cmd)
+      Timeout.timeout(60) do
```
akihikodaki, line 551: Timeout should be performed in the main thread.

lib/setupmanagers/qemuhck/devices/vhost-user-fs-pci.json
```diff
@@ -12,7 +12,8 @@
   "pre_start_commands": [
     "mkdir -p @fs_daemon_share_path@",
     "rm -f @fs_daemon_socket@",
-    "@fs_daemon_bin@ --socket-path=@fs_daemon_socket@ -o source=@fs_daemon_share_path@ -o cache=always &"
+    "@fs_daemon_bin@ --socket-path=@fs_daemon_socket@ -o source=@fs_daemon_share_path@ --cache always &",
+    "bash -c 'until [ -S \"@fs_daemon_socket@\" ]; do echo \"Socket does not exist, waiting...\"; sleep 1; done'"
```
akihikodaki, line 16: I didn't know pre_start_commands are not executed in bash.
It's not good to use a different shell only here so let's revert to `while`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: But run_pre_start_commands is run in a separate thread in the loop. How do we perform timeout from the main thread?

kostyanf14: Why we can't just fail qemu_thread?

lib/setupmanagers/qemuhck/devices/vhost-user-fs-pci.json
kostyanf14: I used the same structure for `while`. We can just drop `bash -c `

akihikodaki: Oh, I see. I have just confirmed `until` is available for the POSIX shell too.

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: > timeout around `AutoHCK::HCKTest::run_and_configure_setup`

This is simple from the code view but difficult to select the proper timeout. `configure_setup` takes about 3-5 min with my NVME SSD but takes about 10-15 with HDD. Also on one Windows insider preview version, I see about 20 min initialization (not hangs on during `Waiting for client #{@name} initialization`). 

Currently, we have 30 min timeout in `AutoHCK::HCKClient::synchronize` (`@cooldown_thread&.join(CLIENT_COOLDOWN_TIMEOUT)`). So timeout around `AutoHCK::HCKTest::run_and_configure_setup` should be the biggest. But this is a bad situation to wait 45 minutes because QEMU can't start.

I think we need some local timeout near QEMU run.

akihikodaki: I see. Let's try the second option then; add code to deliver an exception to the thread spawned `qemu_thread`.

akihikodaki: Exceptions in `qemu_thread` are automatically delivered to the main thread due to `Thread.abort_on_exception` in `bin/auto_hck`.

akihikodaki (approved):

PR #268, opened 2023-11-02
Remove the ID generator from AutoHCK

kostyanf14:
AutoHCK is a CI runner, not a manager, and should not control resources between different runs.

Add command line argument to have the ability to specify an ID that is received from external tools.


YanVugenfirer:
While I agree with the PR message. Who will allocate the IDs in the near future?

kostyanf14:
> While I agree with the PR message. Who will allocate the IDs in the near future?

We don't need this in the near future. Currently, we run only one HCK-CI on the server. If we change this, Jenkins can assign an ID.

kostyanf14:
> While I agree with the PR message. Who will allocate the IDs in the near future?

We can use Jenkins EXECUTOR_NUMBER env to assign id in case of multiple test runs.

akihikodaki (approved):

PR #274, opened 2023-11-20
Improve for UEFI

akihikodaki:


akihikodaki:
b55ec27a5e4da2a93a3093111ab3f1d4b435c7ff needs https://github.com/HCK-CI/HLK-Setup-Scripts/pull/19.

lib/setupmanagers/qemuhck/states.json
```diff
@@ -51,5 +51,12 @@
         "true": {
             "drive_cache_options": ",cache=unsafe"
         }
+    },
+    "fw_type": {
+        "uefi": {
+            "devices_list": [
+                "tpm-tis"
```
kostyanf14, line 58: We need to update the devices_list processing logic in this case.  The current PR has a problem with running SVVP. `tpm-tis` occurs twice in devices_list and auto_hck fails when trying to start the TPM server again.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -39,16 +39,20 @@ def initialize_project(project)
...
+    def studio_vm_options
+      {
+        'id' => @id.to_i,
+        'client_id' => 0,
+        'workspace_path' => @workspace_path,
+        'image_name' => @platform['st_image'],
+        'logger' => @logger,
+        'slirp' => @slirp,
+        'iso_path' => @project.config['iso_path']
+      }.merge(platform_options)
```
kostyanf14, line 51: I am okay with these changes until HLK Studio can successfully be installed with an enabled secure boot. Currently, there is no limitation.

lib/setupmanagers/qemuhck/states.json
akihikodaki: Done with commit e3ee8668b9bdccd10c375f9dfd44c8f4795e1cf1.

kostyanf14 (approved):

PR #291, opened 2024-02-01
Add JSON validation tests

kostyanf14:
This is the initial commit of JSON schemas and validation. For now, allow to have JSONs without schema.

I will send the next PRs with other schemas and changes in the source (if necessary)

### DIFF
--- a/drivers.schema.json
+++ b/drivers.schema.json
@@ -1,0 +1,119 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "patternProperties": {
+    ".*": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "device": {
+          "type": "string"
+        },
+        "inf": {
+          "type": "string"
+        },
+        "sys": {
+          "type": "string"
+        },
+        "install_method": {
+          "type": "string",
+          "enum": [
+            "PNP",
+            "no-drv",
+            "custom"
+          ]
+        },
+        "install_cert": {
+          "type": "boolean"
+        },
+        "install_command": {
+          "type": "string"
+        },
+        "type": {
+          "type": "integer"
+        },
+        "support": {
+          "type": "boolean"
+        },
+        "enlightenments_state": {
+          "type": "boolean"
+        },
+        "pretestcommands": {
+          "type": "array",
+          "items": [
+            {
+              "type": "object",
+              "properties": {
+                "desc": {
+                  "type": "string"
+                },
+                "run": {
+                  "type": "string"
+                }
+              },
+              "required": [
+                "desc",
+                "run"
+              ]
+            }
+          ]
+        },
+        "extra_software": {
+          "type": "array",
+          "items": [
+            {
+              "type": "string"
+            }
+          ]
+        },
+        "reject_test_names": {
+          "type": "array",
+          "items": [
+            {
+              "type": "string"
+            }
+          ]
+        },
+        "select_test_names": {
+          "type": "array",
+          "items": [
+            {
+              "type": "string"
+            }
+          ]
+        }
+      },
+      "if": {
+        "properties": {
+          "install_method": {
+            "const": "custom"
+          }
+        }
+      },
+      "then": {
+        "required": [
+          "install_cert",
+          "install_command"
+        ]
+      },
+      "else": {
+        "not": {
+          "required": [
+            "install_cert",
+            "install_command"
+          ]
+        }
+      },
+      "required": [
+        "name",
+        "device",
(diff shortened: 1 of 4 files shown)


spec/json_validation_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+describe 'json_validation' do
+  Dir['./**/*.json'].each do |json_file|
+    next if json_file.include? 'schema.json'
+
+    name = File.basename(json_file, '.json')
+    dir = File.dirname(json_file)
+    it json_file.to_s do
+      json_schema_candidates = [
+        "#{dir}/#{name}.schema.json",
+        "#{dir}/schema.json",
```
akihikodaki, line 13: Why will you need these two forms?

spec/json_validation_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        raise
+      end
+
+      json_data = JSON.parse(File.read(json_file))
+      json_schema = JSON.parse(File.read(json_schema_files.first))
+
+      schemer = JSONSchemer.schema(json_schema)
+      result = schemer.validate(json_data).to_a
+
+      expect(result).to eq([])
```
akihikodaki, line 28: Simpler:

```suggestion
      result = JSONSchemer.schema(json_schema).validate(json_data)
      expect(result).to contain_exactly()
```

akihikodaki (changes requested):

spec/json_validation_spec.rb
kostyanf14: In the case of platforms Win2022x64, Win2016x64, etc we will have one schema for a lot of files. Same with devices.

spec/json_validation_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        "#{dir}/schema.json",
+      ]
+
+      json_schema_files = json_schema_candidates.select { |f| File.exist?(f) }
+      if json_schema_files.empty?
+        pending("NO SCHEMA FOR #{name} -> #{json_schema_candidates}")
+        raise
+      end
+
+      json_data = JSON.parse(File.read(json_file))
```
akihikodaki, line 22: - Use `Enumerable#filter_map`.
- Rescue `Errno::NOENT` instead of calling `File.exist?` before `File.read`.

spec/json_validation_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      json_schema = JSON.parse(File.read(json_schema_files.first))
+
+      result = JSONSchemer.schema(json_schema).validate(json_data).to_a
```
akihikodaki, line 25: No need for `to_a`.

spec/json_validation_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      ]
+
+      json_schema_files = json_schema_candidates.select { |f| File.exist?(f) }
+      if json_schema_files.empty?
+        pending("NO SCHEMA FOR #{name} -> #{json_schema_candidates}")
+        raise
+      end
+
+      json_data = JSON.parse(File.read(json_file))
+      json_schema = JSON.parse(File.read(json_schema_files.first))
```
akihikodaki, line 23: Give `Pathname.new(json_schema_files.first)` and let `JSONSchemer` read the file.

akihikodaki (changes requested):

spec/json_validation_spec.rb
akihikodaki: Makes sense. I added a few more comments for the logic to handle the two forms.

spec/json_validation_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        File.read(f)
+      rescue Errno::ENOENT
+      end
+
+      if json_schemas.empty?
+        pending("NO SCHEMA FOR #{name} -> #{json_schema_candidates}")
+        raise
+      end
+
+      json_data = JSON.parse(File.read(json_file))
```
akihikodaki, line 26: I rechecked documentation and found `JSON.load_file` so let's use it.

akihikodaki (changes requested):

drivers.schema.json
```diff
@@ -0,0 +1,119 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
akihikodaki, line 2: Sorry for adding a comment again, but I found this is not the latest version. The official document says to use: `https://json-schema.org/draft/2020-12/schema`
https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition

akihikodaki (changes requested):

akihikodaki (approved):

PR #297, opened 2024-02-09
Add PID to each CmdRun log

kostyanf14:
When multiple commands start in parallel it is difficult to understand what exactly fails.


YanVugenfirer (approved):

akihikodaki (changes requested):
Nice improvement but I think printing PID is better as it can be used for `kill`, `gdb`, etc. during debugging and also saves code to generate and keep UUID.

akihikodaki (approved):

PR #298, opened 2024-02-09
Add platforms schema.json

kostyanf14:


lib/engines/hcktest/platforms/schema.json
```diff
@@ -0,0 +1,130 @@
...
+    },
+    "enlightenments_state": {
+      "type": "boolean"
+    },
+    "device_extra": {
+      "type": "string"
+    },
+    "st_image": {
+      "type": "string"
+    },
```
akihikodaki, line 44: Shouldn't these be moved into the `if`? Otherwise a physical platform can specify e.g., `q35` for `machine_type`, but it's clearly wrong.

lib/engines/hcktest/platforms/schema.json
```diff
@@ -0,0 +1,130 @@
...
+    "clients": {
+      "type": "object",
+      "patternProperties": {
+        "c[0-9]": {
```
akihikodaki, line 48: Why do we need this constraint? Isn't any name OK?

lib/engines/hcktest/platforms/schema.json
```diff
@@ -0,0 +1,130 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "kit": {
+      "type": "string"
+    },
+    "type": {
```
akihikodaki, line 11: The name `type` is confusing. we already have `machine_type` and `fw_type` so type of what?

We already call `setupmanager` the concept abstracting QemuHCK and PhysHCK, so it's better to follow the convention and name it `setup` also for consistency. We can name it `setup_type` for consistency with `machine_type` and `fw_type`, but there is `cpu`, which doesn't have `_type` suffix so I think either is fine. Personally I think `_type` suffix is verbose.

lib/engines/hcktest/platforms/schema.json
```diff
@@ -0,0 +1,130 @@
...
+      "type": "string",
+      "enum": [
+        "virtual",
+        "physical"
```
akihikodaki, line 15: Let's name them `qemuhck` and `physhck` for consistency too. Especially It's better to avoid the name `virtual` as the specified properties are mostly specific to QEMU and we will want to have another type when we introduce support for e.g., Hyper-V.

lib/engines/hcktest/platforms/schema.json
```diff
@@ -0,0 +1,130 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
akihikodaki, line 2: You forgot to update this.

akihikodaki (changes requested):
Having a schema is nice addition, but honestly I'm surprised by the verbosity of JSON Schema. Isn't there a better alternative? A quick Web search showed JSON Type Definition and it's much less verbose. I'm open to other options too.
https://jsontypedef.com/

spec/json_validation_spec.rb
```diff
@@ -24,9 +25,10 @@
       end
 
       json_data = JSON.load_file(json_file)
+      schema = JTD::Schema.from_hash(json_schemas.first)
```
akihikodaki, line 28: Call `schema.verify` to ensure the correctness.

lib/engines/hcktest/platforms/jtd.json
```diff
@@ -0,0 +1,59 @@
...
+      "properties": {
+        "name": { "type": "string" },
+        "kit": { "type": "string" },
+        "clients": {
+          "properties": {
+            "c1": { "ref": "physhck_client" }
+          },
+          "optionalProperties": {
+            "c2": { "ref": "physhck_client" }
+          }
```
akihikodaki, line 14: hcktest assumes there are two clients so `PhyWin2016.json` and `PhyWin2019.json` should be fixed instead.

lib/engines/hcktest/drivers/jtd.json
```diff
@@ -0,0 +1,28 @@
+{
+  "definitions": {
+    "command_info": {
+      "properties": {
+        "desc": { "type": "string" },
+        "run": { "type": "string" }
+      }
+    }
+  },
```
akihikodaki, line 9: It's fine but I wonder why you make the use of definition here.

akihikodaki (changes requested):

lib/engines/hcktest/platforms/jtd.json
kostyanf14: hcktest should not assume this. Different testing scenarios expect different machine counts. Our physical HLK has only one client but for the Server Stress SVVP test, we need three clients.

akihikodaki: It makes sense.

lib/engines/hcktest/drivers/jtd.json
kostyanf14: Just copy-paste from the platform.  Do you want to remove this?

akihikodaki: It makes sense. Let's leave it.

spec/json_validation_spec.rb
kostyanf14: This makes sense. One more question. Should I format JTD or leave it as is now?

akihikodaki: It's fine to leave it as is for now. Ideally, we should have the CI enforce some format though.

akihikodaki (approved):

PR #300, opened 2024-02-12
Fix "time_to_second" helper

Jedoku:


lib/auxiliary/time_helper.rb
```diff
@@ -4,9 +4,11 @@
 module AutoHCK
```
kostyanf14, line 4: `Signed-off-by: Vitalii Chulak <vchulakv@gmail.com>`: use @daynix mail there

`fix time to second helper`:  commit title starts with an uppercase letter
`time_to_second` is a function, so better to write its name as is in the commit message without any splitting

kostyanf14 (changes requested):

lib/auxiliary/time_helper.rb
```diff
@@ -4,9 +4,11 @@
...
   # Helper module
   module Helper
-    def time_to_seconds(time)
-      time.split(':').reverse.map.with_index { |a, i| a.to_i * (60**i) }
-          .reduce(:+)
+    def time_to_seconds(time_str)
+      days, time_part = time_str.split('.')
```
akihikodaki, line 8:
```suggestion
      days, time_part = time_str.split('.', 2)
```

Looking at .NET's own parser:
https://github.com/dotnet/runtime/blob/v8.0.1/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Utf8Parser/Utf8Parser.TimeSpan.C.cs

There are the following possible formats:
- dd
- hh:mm
- dd.hh:mm
- hh:mm:ss
- dd.hh:mm:ss
- hh:mm:ss.fffffff
- dd.hh:mm:ss.fffffff

As shown here, `.` may appear before the second fraction too. We don't want to split it.

Also let's name `days` as `days_str` to make it consistent with `time_str` and to make it clear it is a string unlike `hours`, `minutes` and `seconds`.

lib/auxiliary/time_helper.rb
```diff
@@ -4,9 +4,11 @@
...
+      hours, minutes, seconds = time_part.split(':').map(&:to_i)
+      total_hours = (days.to_i * 24) + hours
+      (total_hours * 3600) + (minutes * 60) + seconds
```
akihikodaki, line 11: Why don't you use the original code for `time_part` parsing?

lib/auxiliary/time_helper.rb
```diff
@@ -4,9 +4,11 @@
 module AutoHCK
   # Helper module
   module Helper
-    def time_to_seconds(time)
-      time.split(':').reverse.map.with_index { |a, i| a.to_i * (60**i) }
-          .reduce(:+)
+    def time_to_seconds(time_str)
+      days, time_part = time_str.split('.')
+      hours, minutes, seconds = time_part.split(':').map(&:to_i)
```
akihikodaki, line 9: Use `to_f` to parse the second fraction.

akihikodaki (changes requested):
Please add tests if you are not in a hurry. Use the test cases from .NET:
https://github.com/dotnet/runtime/blob/v8.0.1/src/libraries/System.Memory/tests/ParsersAndFormatters/Parser/TestData.Parser.TimeSpan.cs#L181-L206

You don't need to test the cases of leading whitespaces, negative values, and illegal cases. Such cases should never happen.

lib/auxiliary/time_helper.rb
kostyanf14: @akihikodaki I am ok with adding tests, just a question. Do we need to think about these formats:

- dd
- hh:mm
- dd.hh:mm
- hh:mm:ss.fffffff
- dd.hh:mm:ss.fffffff

All formats looks ok for MS, but I don't think HLK will provide something in its. Or at least ignore `second fraction`.

akihikodaki: We will never have the following according to the documentation https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings and source code https://github.com/dotnet/runtime/blob/v8.0.1/src/libraries/System.Private.CoreLib/src/System/Globalization/TimeSpanFormat.cs
- dd
- hh:mm
- dd.hh:mm

S second fraction is still possible. I think it's easy to convert to use a floating point number or just to ignore though (`to_i` will ignore the fraction).

spec/lib/auxiliary/test_helper_spec.rb
```diff
@@ -0,0 +1,23 @@
+require_relative '../../../lib/auxiliary/time_helper.rb'
```
kostyanf14, line 1: `s/test_helper_spec.rb/time_helper_spec.rb/`

spec/lib/auxiliary/test_helper_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+RSpec.describe AutoHCK::Helper do
+  include AutoHCK::Helper
+
+  describe '#time_to_seconds' do
+    
+    time_strings = {
+      "1:2" => 3720, #hh:mm
```
kostyanf14, line 9: I don't prefer to have magic numbers without comments. 

You wrote `"1.2:3:4" => 86400 + 7380 + 4`, then I preferred `"1.2:3:4" => 86400 + 2 * 60 * 60 + 3 * 60 + 4` 
Other variants add comments before `time_strings` that the number is a second count and wrote `93784` instead of  `86400 + 7380 + 4`

Currently, it looks strange why we sometimes calculate something but sometimes have a constant.

spec/lib/auxiliary/test_helper_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      "1:2:3" => 3723, #hh:mm:ss
+      "1.2:3" => 86400 + 7380, #dd.hh:mm
+      "1.2:3:4" => 86400 + 7380 + 4, #dd.hh:mm:ss
+      "1:2:3.22" => 3723.22, #hh:mm:ss.ff
+      "1.2:3:4.22" => 86400 + 7380 + 4.22, #dd.hh:mm:ss.ff
+    }
+
+    time_strings.each do |time, expected|
```
kostyanf14, line 17: what is `expected`?  better `|time_string, seconds|`

lib/auxiliary/time_helper.rb
```diff
@@ -4,9 +4,18 @@
 module AutoHCK
   # Helper module
   module Helper
-    def time_to_seconds(time)
-      time.split(':').reverse.map.with_index { |a, i| a.to_i * (60**i) }
-          .reduce(:+)
+    def time_to_seconds(time_string)
+      match = time_string.match(/^(?:(\d+)\.)?(?:(\d+):)?(\d+)(?::(\d+))?(?:\.(\d+))?$/)
```
akihikodaki, line 8: To match entire second including fraction:

```suggestion
      match = time_string.match(/^(?:(\d+)\.)?(?:(\d+):)?(\d+)(?::(\d+(?:\.\d+)?))?$/)
```

spec/lib/auxiliary/test_helper_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require_relative '../../../lib/auxiliary/time_helper'
+
+RSpec.describe AutoHCK::Helper do
+  include AutoHCK::Helper
+
+  describe '#time_to_seconds' do
+    time_strings = {
```
akihikodaki, line 9: It is a bit awkward to name `time_strings` as it doesn't only have strings but also numbers. They are test cases so we can call `cases` or something like that.

spec/lib/auxiliary/test_helper_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require_relative '../../../lib/auxiliary/time_helper'
+
+RSpec.describe AutoHCK::Helper do
```
akihikodaki, line 5:
```suggestion
describe AutoHCK::Helper do
```

For consistency with other files. Not sure what this difference implies though.

akihikodaki (changes requested):

lib/auxiliary/time_helper.rb
Jedoku: i believe that code was more explicit and easier to understand at a glance

kostyanf14:
@Jedoku LGTM

Please combine all your changes into one or two commits. For each review iteration, you should fix the original commits.

kostyanf14:
@Jedoku You missed `Signed-off-by` in commit

kostyanf14 (approved):

PR #303, opened 2024-02-15
AutoHCK fail when rerun test with BSOD

Jedoku:
Resolves: #254

### DIFF
--- a/lib/auxiliary/time_helper.rb
+++ b/lib/auxiliary/time_helper.rb
@@ -24,5 +24,9 @@ module AutoHCK
     def time_diff(start, finish)
       ((finish - start) * 24 * 60 * 60).to_i
     end
+
+    def current_timestamp
+      Time.now.strftime('%Y_%m_%d_%H_%M_%S')
+    end
   end
 end
--- a/lib/engines/hcktest/tests.rb
+++ b/lib/engines/hcktest/tests.rb
@@ -289,7 +289,7 @@ module AutoHCK
       status_count('InQueue').zero? && current_test.nil?
     end
 
-    def collect_memory_dump(machine, l_tmp_path)
+    def download_memory_dump(machine, l_tmp_path)
       exist = @tools.exists_on_machine?(machine, '${env:SystemRoot}/Minidump')
       @logger.debug("Checking Minidump exist on #{machine}: #{exist}")
       return false unless exist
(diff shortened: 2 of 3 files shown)


lib/engines/hcktest/tests.rb
```diff
@@ -307,7 +307,7 @@ def collect_memory_dumps(test)
       l_zip_path = "#{@project.workspace_path}/memory_dump_#{id}.zip"
       l_tmp_path = "#{@project.workspace_path}/tmp_#{id}"
 
-      collected_client = collect_memory_dump(@client.name, "#{l_tmp_path}/#{@client.name}")
+      collected_client = collect_memory_dump(@client.name, "#{l_tmp_path}/#{@client.name}_#{timestamp}")
       collected_support = collect_memory_dump(@support.name, "#{l_tmp_path}/#{@support.name}") unless @support.nil?
```
kostyanf14, line 311: Both clients can generate BSOD. So add timestamp for support VM too

kostyanf14 (changes requested):

akihikodaki (changes requested):
Can you have an explanation for me, who doesn't know the context? For me, it looks immune against collisions even without a timestamp since the zip has the id.

Why did you use id instead of timestamp? It's inconsistent.
If you really need timestsamps, extract `AutoHCK::Project#create_timestamp` into a helper and change this to use it.

Jedoku:
> Can you have an explanation for me, who doesn't know the context? For me, it looks immune against collisions even without a timestamp since the zip has the id.
> 
> Why did you use id instead of timestamp? It's inconsistent. If you really need timestsamps, extract `AutoHCK::Project#create_timestamp` into a helper and change this to use it.

When we start a test with a dump twice, we have to save both dumps into a zip archive. The problem arises when we save a folder with the name of the client's VM (CL01 in this case) into the zip file. When we try to add the folder a second time, we receive an error stating that the folder already exists. The solution to this issue involves adding a timestamp to the folder name, allowing us to circumvent the problem.

akihikodaki:
@Jedoku I see. Thanks for the explanation.

lib/auxiliary/time_helper.rb
```diff
@@ -26,7 +26,7 @@ def time_diff(start, finish)
     end
 
     def timestamp
```
akihikodaki, line 28: Naming this timestamp will collide with `Project#timestamp` attr_reader so it's better to avoid.

akihikodaki (changes requested):

lib/auxiliary/time_helper.rb
```diff
@@ -24,5 +24,9 @@ def seconds_to_time(sec)
     def time_diff(start, finish)
       ((finish - start) * 24 * 60 * 60).to_i
     end
+
+    def add_timestamp
```
kostyanf14, line 28: This function does not add a timestamp it gets a timestamp. As `get_timestamp` is a bad name from a Ruby point of view, use `create_timestamp`, `generate_timestamp`, `current_timestamp`, etc

kostyanf14 (changes requested):

kostyanf14 (approved):

lib/engines/hcktest/tests.rb
```diff
@@ -301,16 +301,23 @@ def collect_memory_dump(machine, l_tmp_path)
       true
     end
 
+    def download_memory_dumps(l_tmp_path)
+      collected_client = download_memory_dump(@client.name, "#{l_tmp_path}/#{@client.name}_#{current_timestamp}")
```
akihikodaki, line 305: We only download dumps here; please rename `collected_client` to `downloaded_client`. Do the same for `collected_support` too.

akihikodaki (changes requested):

akihikodaki (approved):

PR #308, opened 2024-02-20
Specify platform memory as number in GB

kostyanf14:


akihikodaki (approved):
Now it's validated by JTD too :+1:

kostyanf14:
> Now it's validated by JTD too 👍

Yes, need the same for "cpus",  "winrm_port"

PR #310, opened 2024-02-22
RFC: Sorbet integration

kostyanf14:
Please ignore everything from the `sorbet` folder during a review.


lib/engines/hckinstall/hckinstall.rb
```diff
@@ -363,9 +367,10 @@ def prepare_studio_drives
     def copy_drivers
       @logger.info('HCKInstall: Copy all drivers')
       FileUtils.rm_rf("#{@hck_setup_scripts_path}/drivers")
+      # (src, dest, preserve = false, dereference_root = false, remove_destination = false);
       FileUtils.copy_entry(@project.options.install.driver_path,
                            "#{@hck_setup_scripts_path}/drivers",
-                           remove_destination: true)
+                           false, false, true)
```
akihikodaki, line 373: First bug spotted by Sorbet :+1:

By the way, why would we want `remove_destination` when we remove the whole directory just before that?

lib/engines/hcktest/hcktest.rb
```diff
@@ -197,7 +200,7 @@ def run_and_configure_setup(scope)
       end
     rescue AutoHCKError => e
       @project.logger.warn("Running and configuring setup failed: (#{e.class}) #{e.message}")
-      raise e unless (retries += 1) < AUTOHCK_RETRIES
+      raise e unless (retries = T.must(retries) + 1) < AUTOHCK_RETRIES
```
akihikodaki, line 203: We should `begin` a block after `retries` instead.

.github/workflows/check.yml
```diff
@@ -62,6 +62,34 @@ jobs:
...
+      with:
+        ruby-version: 3.1.4
+        bundler-cache: false
+
+    - name: Install dependencies
+      run: bundle install
+
+    - name: Run sorbet type check
+      run: bundle exec srb typecheck
+
```
akihikodaki, line 92: We already have different jobs for `rspec` and `rubocop`, but they actually should be one job; we can save many lines and computation time for setup in that way.

akihikodaki (changes requested):
It doesn't make sense to check in RBIs when you don't check in gems.

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      data = nil
+      if json_file.include? 'lib/engines/hcktest/drivers'
+        data = AutoHCK::Models::Driver.from_hash(json_data)
+      else
+        pending("NO TYPE FOR #{json_file}")
+        raise
+      end
+
+      expect(data).not_to eq(nil)
```
akihikodaki, line 20:
```suggestion
      expect(data).not_to be_nil
```

akihikodaki:
Thanks for providing a proof of concept. This is very informative.

Having two definitions for one JSON is terrible so we should pick one. I have mixed feelings for Sorbet, but I think the utility of its static type checker supasses its disadvantages.

We can write a converter to Sorbet if we want to write a schema in another language, but it's hard to support Sorbet's static type checker with such a converter. Supporting its runtime type check-ing is easy since we can generate a type definition as Ruby objects on Ruby, but static type checking requires the generated definitions to be saved as Ruby code, and that's very complicated.

The other way around is easy in contrary. If we will have a frontend written in JavaScript and need to consume Sorbet code, we can traverse the types and generate JSON, which is only a matter of calling `JSON.dump`.

Honestly, I don't like some syntactical aspects of Sorbet; it requires typing `T::` a lot, and the enum idiom does not make sense, but I think we can tolerate that.

lib/engines/hckinstall/hckinstall.rb
akihikodaki: Opened: https://github.com/ruby/ruby/pull/10081

Gemfile
```diff
@@ -15,8 +15,18 @@ gem 'openssl', require: false
 gem 'rtoolsHCK', git: 'https://github.com/HCK-CI/rtoolsHCK.git', tag: 'v0.4.0'
 gem 'rubyzip'
 gem 'sentry-ruby'
+gem 'sorbet-runtime'
 gem 'sqlite3'
 
+group :development do
+  gem 'sorbet'
+  gem 'webrick'
```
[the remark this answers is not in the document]
kostyanf14: For some reason `srb init` and `tapioca init` fail with the error no `webrick`. As this is RFC, I don't investigate.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -174,14 +177,15 @@ def drivers
 
       @need_copy_drivers = false
       drivers.each do |driver|
-        next if driver['install_method'] == 'no-drv'
+        next if driver.install_method == Models::DriverInstallMethods::NoDrv
```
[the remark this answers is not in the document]
kostyanf14: We can add functions to the `driver` model, we can use `include`, `extend` to make less verbosity.

bin/tapioca
```diff
@@ -0,0 +1,27 @@
...
+  else
+    abort("Your `bin/bundle` was not generated by Bundler, so this binstub cannot run.
+Replace `bin/bundle` by running `bundle binstubs bundler --force`, then run this command again.")
+  end
+end
+
+require 'rubygems'
+require 'bundler/setup'
+
+load Gem.bin_path('tapioca', 'tapioca')
```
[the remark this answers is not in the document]
kostyanf14: Just autogenerated. Of course, we can remove and I also don't use `bin/` files.

lib/engines/hckinstall/hckinstall.rb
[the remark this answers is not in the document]
akihikodaki: Perhaps we can have `AutoHCK::Models::Base` which inherits `T::Struct` and has `read_json`/`from_json`/whatever.

bin/tapioca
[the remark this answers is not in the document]
akihikodaki: Let's remove it; I wonder why it doesn't autogenerate for `srb`.

.github/workflows/check.yml
akihikodaki: Add `continue-on-error` for each of typecheck, rspec, and rubocop. You can later check if all of the steps succeeded.

kostyanf14:
@akihikodaki I don't touch workflow-related commit, I want to investigate the ability to simplify it. As this is not related to Sorbet in general, let's move the deduplication part outside this PR.

lib/auxiliary/json_helper.rb
```diff
@@ -12,13 +14,16 @@ module AutoHCK
...
+      extend T::Sig
+
       @json_override_file = 'override.json'
 
       def self.update_json_override(json_file)
         @json_override_file = json_file
       end
 
-      def self.read_json(json_file, logger)
+      sig { params(json_file: String, logger: T.nilable(T.class_of(::Logger))).returns(T::Hash[String, T.untyped]) }
```
akihikodaki, line 25:
```suggestion
      sig { params(json_file: String, logger: T.nilable(::Logger)).returns(T::Hash[String, T.untyped]) }
```

sorbet/config
```diff
@@ -0,0 +1,4 @@
+--dir
+.
+--ignore=tmp/
+--ignore=vendor/
```
akihikodaki, line 4: I'm not sure about these configurations. There is no `tmp/` and `vendor/` in my setup.

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'json'
+
+require './lib/models/driver'
+
+describe 'srb_json_load' do
+  Dir['./**/*.json'].each do |json_file|
+    next if json_file.include? 'jtd.json'
+
+    it json_file.to_s do
+
```
akihikodaki, line 10: Unnecessary blank line.

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      data = nil
+      if json_file.include? 'lib/engines/hcktest/drivers'
+        data = AutoHCK::Models::Driver.from_json_file(json_file)
+      else
+        pending("NO TYPE FOR #{json_file}")
+        raise
+      end
```
akihikodaki, line 17:
```suggestion
      unless json_file.include? 'lib/engines/hcktest/drivers'
        pending("NO TYPE FOR #{json_file}")
        raise
      end

      data = AutoHCK::Models::Driver.from_json_file(json_file)
```

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -59,7 +59,7 @@ def drivers_options
       options = {}
       @drivers&.each do |driver|
         OPT_NAMES.each do |name|
-          options[name] = driver[name] unless driver[name].nil?
+          options[name] = driver.instance_variable_get("@#{name}") unless driver.instance_variable_get("@#{name}").nil?
```
akihikodaki, line 62: Use `public_send`. It's not nice to read instance variables, which are inherently private, from outside.

akihikodaki (changes requested):

akihikodaki:
> I think these changes are good enough to start with Sorbet.
> 
> If there are no critical problems that we can see now, let's continue in the next way:
> 
>     1. merge this PR
> 
>     2. remove JTD related code
> 
>     3. add models for other objects that we get from JSON
> 
>     4. enable typing in other parts of the source

That sounds good to me.

sorbet/config
kostyanf14: We use `BUNDLE_PATH: "vendor/bundle"` in the installer. `tmp` is redundant there.

spec/srb_json_load_spec.rb
kostyanf14: I am not sure about this. In the next commit, we will add platforms model and add `if json_file.include? 'lib/engines/hcktest/platforms'` or do you have another idea of how to select the proper type for JSON?

akihikodaki: What about having an array typed `T::Array[[String, AutoHCK::Models::JsonHelper]]`, where the `String` is a glob pattern and `AutoHCK::Models::JsonHelper` is the corresponding model?

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'json'
+
+require './lib/models/driver'
+
+JSON_TYPES_MAP = {
+  './lib/engines/hcktest/drivers/*.json' => AutoHCK::Models::Driver,
+}.freeze
+
+describe 'srb_json_load' do
+  Dir['./**/*.json'].each do |json_file|
```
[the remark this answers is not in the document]
kostyanf14: I want to check every JSON that we have. When we will remove `pending` I expect to have `rspec` fail if only JSON was added without type or exclusion (for spec for example).

kostyanf14:
> I think we should also run `tapioca gems --verify`. Their documentation says it breaks Dependabot, but it apparently does not. `dependabot-core` itself uses Dependabot with Sorbet and it seems just fine: https://github.com/dependabot/dependabot-core/issues?page=3&q=label%3Adependencies+label%3Aruby+is%3Aclosed

I think they mean this problem https://github.com/dependabot/dependabot-core/issues/5962. Need to download tapioca definitions after updating a gem.

akihikodaki:
> > I think we should also run `tapioca gems --verify`. Their documentation says it breaks Dependabot, but it apparently does not. `dependabot-core` itself uses Dependabot with Sorbet and it seems just fine: https://github.com/dependabot/dependabot-core/issues?page=3&q=label%3Adependencies+label%3Aruby+is%3Aclosed
> 
> I think they mean this problem [dependabot/dependabot-core#5962](https://github.com/dependabot/dependabot-core/issues/5962). Need to download tapioca definitions after updating a gem.

Yes, what I don't get is that https://github.com/dependabot/dependabot-core/ doesn't have such code yet it seems working.

lib/auxiliary/json_helper.rb
```diff
@@ -1,24 +1,32 @@
...
+
       @json_override_file = 'override.json'
 
       def self.update_json_override(json_file)
         @json_override_file = json_file
       end
 
-      def self.read_json(json_file, logger)
+      sig do
+        params(json_file: String, logger: T.nilable(T.any(MultiLogger, ::Logger))).returns(T::Hash[String, T.untyped])
```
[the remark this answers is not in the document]
kostyanf14: >Will it take ::Logger?

When we move this part to gem for a separate project. I think yes.

>Revisiting this, I also wonder if taking nil makes sense.

These are simplify tests for example.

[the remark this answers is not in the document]
akihikodaki: It makes sense.

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    next if json_file.include? 'jtd.json'
+
+    it json_file.to_s do
+      types = JSON_TYPES_MAP.filter_map { |json_files_pattern, type| type if File.fnmatch(json_files_pattern, json_file) }
```
akihikodaki, line 14:
```suggestion
      type = JSON_TYPES_MAP.find { File.fnmatch(_1, json_file) }[1]
```

So that you don't need to call `first` later.

spec/srb_json_load_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      if types.empty?
+        pending("NO TYPE FOR #{json_file}")
+        raise
+      end
+
+      data = types.first.from_json_file(json_file)
+      expect(data).not_to be_nil
```
akihikodaki, line 22: This `expect` shouldn't be needed; `from_json_file` always returns non-`nil` value according to its signature.

akihikodaki (changes requested):
Things I have pointed out earlier are all correctly resolved. I still have a few comments though.

spec/srb_json_load_spec.rb
kostyanf14: but we should expect that there is no exception. Is it correct to have `it` without any `expect`?

Gemfile
```diff
@@ -15,8 +15,17 @@ gem 'openssl', require: false
...
 gem 'sqlite3'
 
+group :development do
+  gem 'sorbet'
+end
+
+group :development, :test do
+  gem 'tapioca', '= 0.12.0', require: false
+end
+
```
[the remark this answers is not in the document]
kostyanf14: `require: false` and `group` copy-pasted from https://sorbet.org/docs/adopting#step-1-install-dependencies
`version` was committed by mistake.

spec/srb_json_load_spec.rb
kostyanf14: for all JSONs without type:
```
srb_json_load ./svvp.json
      Failure/Error: type = JSON_TYPES_MAP.find { |json_files_pattern, type| File.fnmatch(json_files_pattern, json_file) }[1]
      
      NoMethodError:
        undefined method `[]' for nil:NilClass
```

Gemfile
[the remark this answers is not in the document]
akihikodaki: Regarding `group`, I'm fine either adding `sorbet` into `test` or excluding `tapioca` from `test`. But they should be consistent.
It is probably a good idea to add `require: false` to `sorbet`; we only use the command line.

spec/srb_json_load_spec.rb
akihikodaki: OK. Fix like:

```suggestion
      pair = JSON_TYPES_MAP.find { File.fnmatch(_1[0], json_file) }
```

The idea is to use `find` to remove `first` call.

spec/srb_json_load_spec.rb
akihikodaki: It should be fine. It is also possible to write:
```Ruby
expect { types.first.from_json_file(json_file) }.not_to raise_error
```

Jedoku (approved):

PR #319, opened 2024-03-06
QemuHCK: Do not add existing values from states to array

kostyanf14:


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -300,7 +300,7 @@
         when Hash
           var_value.merge! value
         when Array
-          var_value << value unless var_value.include? value
+          var_value |= value
```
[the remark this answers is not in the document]
kostyanf14: @akihikodaki I think this is false-positive, `var_value` is not a local variable in our case.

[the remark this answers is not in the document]
akihikodaki: This is not a false positive; you need to `instance_variable_set`.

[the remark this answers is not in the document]
kostyanf14: @akihikodaki Fixed

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -220,7 +220,7 @@ def define_local_variables
       @machine_extra_param = []
       @device_extra_param = []
       @iommu_device_param = []
-      @cpu_options = []
+      @cpu_options = ['@cpu@', '+x2apic', '+fsgsbase', 'model=@cpu_model@']
```
akihikodaki, line 223: I think it's better not to put `@cpu@` here; `@cpu@` is different from other options in terms that it must be present before the other options.

Also, here is some Ruby magic:
```suggestion
      @cpu_options = %w[@cpu@ +x2apic +fsgsbase model=@cpu_model@]
```

lib/setupmanagers/qemuhck/states.json
```diff
@@ -49,7 +54,9 @@
     },
     "drive_unsafe_cache_state": {
         "true": {
-            "drive_cache_options": ",cache=unsafe"
+            "drive_cache_options": [
+                ",cache=unsafe"
```
akihikodaki, line 58:
```suggestion
                "cache=unsafe"
```

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/states.json
kostyanf14: Unfortunately, no. We use `@drive_cache_options@` in `"-drive file=@image_path@,if=none,id=ide_@run_id@_@client_id@@drive_cache_options@"` and should somehow add `,` before drive_cache_options if it present.

kostyanf14: In any case we need some if/map/etc to add ',' before all options but only if options is not empty

akihikodaki (approved):

PR #322, opened 2024-03-14
Drop JSON Type Definition

kostyanf14:
JTD was replaced with Sorbet.


akihikodaki (changes requested):
Please remove check for `jtd.json` from `spec/srb_json_load_spec.rb` as well.

akihikodaki (approved):

YanVugenfirer (approved):

PR #325, opened 2024-03-18
Support HCK installer as ISO

Jedoku:
Script side https://github.com/HCK-CI/HLK-Setup-Scripts/pull/20


lib/engines/hckinstall/hckinstall.rb
```diff
@@ -88,11 +87,10 @@ def init_config
     end
 
     def studio_platform(kit)
-      studio_platform_list = Json.read_json(STUDIO_PLATFORM_JSON, @logger)
+      res = read_kit(kit)
```
akihikodaki, line 90: `read_kit` is called twice; it is also called in `init_class_variables`.

lib/engines/hckinstall/setup_scripts_helper.rb
```diff
@@ -23,9 +23,15 @@ def validate_setup_scripts_config(config)
       raise(AutoHCKError, "Undefined HLK setup scripts configs: #{extra_keys.join(', ')}.")
     end
 
-    def download_kit_installer(url, kit, hck_setup_scripts_path)
+    def download_kit_installer(url, kit, hck_setup_scripts_path, iso)
       dw = Downloader.new(@logger)
-      dw.download(url, "#{hck_setup_scripts_path}/Kits/#{kit}Setup.exe")
+      if iso
+        iso_path = "#{hck_setup_scripts_path}/Kits/#{kit}Setup.iso"
+        dw.download(url, iso_path) unless File.exist?(iso_path)
```
akihikodaki, line 30: Why do you need to check `File.exist?(iso_path)` when you don't if `iso` is false?

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -282,13 +280,14 @@ def prepare_setup_scripts_config
...
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+        @kit_path = download_kit_installer(@kit_info['download_url'],
+                                           "#{kit_type}#{kit_version}", @hck_setup_scripts_path, iso: @kit_info['iso'])
       end
 
       installers = [
         "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}Setup.exe",
-        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}/#{kit_type}Setup.exe"
+        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}/#{kit_type}Setup.exe",
+        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}Setup.iso"
```
akihikodaki, line 290: It's not right to add `"#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}Setup.iso"` unconditionally. It will result in an incorrect behavior if `@kit_info['iso']` is true and `@kit_info['download_url']` is `nil`.

akihikodaki (changes requested):
Instead of requiring to specify `iso` flag, why not automatically detect it?
If `File.read(path, 2) == 'MZ'`, it is a Windows executable. You can assume a disk image otherwise.

lib/engines/hckinstall/setup_scripts_helper.rb
kostyanf14: We don't want to download a big ISO every time. If the installation fails near after the ISO download we will spend a lot of time just for downloading. HLK online installer (exe) is several MB, so redownloading it is no so a long process.

lib/engines/hckinstall/hckinstall.rb
kostyanf14: I expect `@kit_info['iso'] is true` and `@kit_info['download_url'] is nil` in production with Insider Preview testing. Microsoft provides temporary URLs, but we can rerun the installation several times due to a problem with Insider Preview Windows / HLK. 

I expect that the HLKInstall engine will see ISO and be as happy as currently with EXE.
```
unless @kit_info['download_url'].nil?
        download_kit_installer(@kit_info['download_url'],
                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
      end
````

We download the installer only if download_url is present, but if I manually extract HLK ISO and `"#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}/#{kit_type}Setup.exe"` will be present, then HLKInstall engine will continue work. With ISO I expect the same behavior without download_url.

File: lib/engines/hckinstall/setup_scripts_helper.rb (hunk 2 above)
Comment (akihikodaki) [reply lib/engines/hckinstall/setup_scripts_helper.rb:30]: `prepare_setup_scripts_config` already has a `File.exist?` check so let's move it before `download_kit_installer` to see if HLK is downloaded, whether it is iso or not.

**Comment by Jedoku** [comment]
> Instead of requiring to specify `iso` flag, why not automatically detect it? If `File.read(path, 2) == 'MZ'`, it is a Windows executable. You can assume a disk image otherwise.

Yeah, we think about it, but for get extention you need to download file from link and check it then somehow save it to another folder  it was easiest way to done this task. We use curb gem for done this and with provided links it`s impossible to get filename before downloading and you also should save it with filename.

**Comment by akihikodaki** [comment]
> > Instead of requiring to specify `iso` flag, why not automatically detect it? If `File.read(path, 2) == 'MZ'`, it is a Windows executable. You can assume a disk image otherwise.
> 
> Yeah, we think about it, but for get extention you need to download file from link and check it then somehow save it to another folder it was easiest way to done this task. We use curb gem for done this and with provided links it`s impossible to get filename before downloading and you also should save it with filename.

Just download it without an extension and rename it.

**Review by akihikodaki — CHANGES_REQUESTED** [review]
- It's better to use `httpclient` or `faraday` gem if you are going to read HTTP headers. Note that these gems are already used by other dependencies, so requiring them does not add more dependencies.
- Don't perform HTTP requests twice. I bet it doesn't need much code to save one HTTP request.

File: lib/engines/hckinstall/hckinstall.rb [hunk 4]
```diff
@@ -235,12 +234,11 @@ def run_first(studio:, client:)
         if studio
           prepare_studio_drives
 
+          iso_list = [@setup_studio_iso, @studio_iso_info['path']]
+          iso_list.push(@kit_path) if @kit_is_iso
```
akihikodaki, line 238: For consistency with the following line:
```suggestion
          iso_list << @kit_path if @kit_is_iso
```

akihikodaki:
Also see: https://github.com/HCK-CI/AutoHCK/pull/325#discussion_r1529613778

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -271,31 +269,49 @@ def run_second(client:)
...
+
+      if @kit_info['download_url'].nil? || kit_present
+        @logger.info("HLK installer #{kit_type}#{kit_version} already exists, download skipped") if kit_present
+        @logger.info('HLK installer download URL is not provided') if @kit_info['download_url'].nil?
+      else
         download_kit_installer(@kit_info['download_url'],
                                "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
       end
 
+      raise(AutoHCKError, "HLK installer #{@kit_path} not found") unless kit_exist?(kit_type, kit_version)
```
akihikodaki, line 289: Don't run `kit_exist?` twice when `@kit_info['download_url'].nil? || kit_present`.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -271,31 +269,49 @@ def run_second(client:)
...
+    end
+
+    def parse_kit_info
+      kit_string = @platform['kit']
+      kit_type = kit_string[0..2]
+      kit_version = kit_type == 'HCK' ? '' : kit_string[3..]
+      [kit_type, kit_version]
+    end
+
+    def kit_exist?(kit_type, kit_version)
```
akihikodaki, line 303: It looks like it only checks the existence, but it actually sets `@kit_path` and `@kit_is_iso`. I suggest renaming this `find_kit` and returning the path to the kit. `@kit_path` and `@kit_is_iso` should be set in `prepare_setup_scripts_config`.

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,54 @@ def run_second(client:)
...
+      @kit_path = find_kit(kit_type, kit_version)
+
+      log_kit_status(@kit_path, kit_type, kit_version)
+
+      unless @kit_info['download_url'].nil? || @kit_path
         download_kit_installer(@kit_info['download_url'],
                                "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
       end
 
+      @kit_is_iso = @kit_path.end_with?('.iso')
```
kostyanf14, line 293: Where you set @kit_path after download_kit_installer. Looks like it is always nil.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,54 @@ def run_second(client:)
...
 
-      unless @kit_info['download_url'].nil?
-        download_kit_installer(@kit_info['download_url'],
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      @kit_path = find_kit(kit_type, kit_version)
+
+      log_kit_status(@kit_path, kit_type, kit_version)
+
+      unless @kit_info['download_url'].nil? || @kit_path
+        @kit_path = download_kit_installer(@kit_info['download_url'],
```
akihikodaki, line 289: Don't download when `!@kit_path.nil?`

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
akihikodaki: No it doesn't. No need for change.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,54 @@ def run_second(client:)
...
       }
 
-      unless @kit_info['download_url'].nil?
-        download_kit_installer(@kit_info['download_url'],
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      @kit_path = find_kit(kit_type, kit_version)
+
+      log_kit_status(@kit_path, kit_type, kit_version)
+
+      unless @kit_info['download_url'].nil? || @kit_path
```
akihikodaki, line 288: `@kit_info['download_url'].nil?` is always `true` here when `@kit_path.nil?` so don't check it.

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,54 @@ def run_second(client:)
...
+    def find_kit(kit_type, kit_version)
       installers = [
         "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}Setup.exe",
-        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}/#{kit_type}Setup.exe"
+        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}/#{kit_type}Setup.exe",
+        "#{@hck_setup_scripts_path}/Kits/#{kit_type}#{kit_version}Setup.iso"
       ]
 
-      raise unless (file = installers.find { File.exist? _1 })
+      return nil unless (kit_path = installers.find { File.exist? _1 })
```
akihikodaki, line 314: Just return `installers.find { File.exist? _1 }`

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,54 @@ def run_second(client:)
...
-      @logger.info("HLK installer #{file} was found")
+      kit_path
+    end
 
-      create_setup_scripts_config(@hck_setup_scripts_path, config)
+    def log_kit_status(kit_presense, kit_type, kit_version)
+      if kit_presense
+        @logger.info("HLK installer #{kit_type}#{kit_version} already exists")
+      elsif @kit_info['download_url'].nil?
+        @logger.info('HLK installer download URL is not provided')
```
akihikodaki, line 323: Raise since we can't continue.

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,52 @@ def run_second(client:)
...
-
-      @logger.info("HLK installer #{file} was found")
+      installers.find { File.exist? _1 }
+    end
 
-      create_setup_scripts_config(@hck_setup_scripts_path, config)
+    def log_kit_status(kit_presense, kit_type, kit_version)
+      if kit_presense
+        @logger.info("HLK installer #{kit_type}#{kit_version} already exists")
+      elsif @kit_info['download_url'].nil? && kit_presense.nil?
```
akihikodaki, line 320: - Remove `kit_presense.nil?`; It's `elsif` so it is always true.
- Raising is not about logging so move out from this method.
- presence, not presense.

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,47 @@ def run_second(client:)
...
 
-      unless @kit_info['download_url'].nil?
-        download_kit_installer(@kit_info['download_url'],
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      @kit_path = find_kit(kit_type, kit_version)
+
+      if @kit_path.nil?
+        @kit_path = download_kit_installer(@kit_info['download_url'],
+                                           "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      elsif @kit_info['download_url'].nil?
```
akihikodaki, line 289: You need to check if `@kit_info['download_url'].nil?` before downloading.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,47 @@ def run_second(client:)
...
+        raise(AutoHCKError, 'HLK installer download URL is not provided and installer is not found')
+      else
+        @logger.info("HLK installer #{kit_type}#{kit_version} already exists")
       end
 
+      @kit_is_iso = @kit_path.end_with?('.iso')
+
+      @logger.info("HLK installer #{kit_type}#{kit_version} was found at #{@kit_path}")
+
+      pause(10)
```
akihikodaki: Why pause?

akihikodaki (changes requested):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,33 +274,50 @@ def run_second(client:)
...
       }
 
-      unless @kit_info['download_url'].nil?
-        download_kit_installer(@kit_info['download_url'],
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      @kit_path = find_kit(kit_type, kit_version)
+
+      if @kit_info['download_url'].nil? && @kit_path.nil?
+        raise(AutoHCKError, 'HLK installer download URL is not provided and installer is not found')
       end
```
akihikodaki, line 288: Why don't you raise in the following `if @kit_path.nil?`? This saves checking `@kit_path.nil?` twice.

akihikodaki (changes requested):

akihikodaki (approved):

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -276,31 +274,48 @@ def run_second(client:)
...
       }
 
-      unless @kit_info['download_url'].nil?
-        download_kit_installer(@kit_info['download_url'],
-                               "#{kit_type}#{kit_version}", @hck_setup_scripts_path)
+      @kit_path = find_kit(kit_type, kit_version)
+
+      if @kit_path.nil?
+        if @kit_info['download_url'].nil?
+          raise(AutoHCKError, 'HLK installer download URL is not provided and installer is not found')
```
kostyanf14, line 288: Let's use move-specific EngineError

kostyanf14 (approved):

PR #333, opened 2024-03-27
qemuhck: Isolate swtpm_localca state

akihikodaki:
The default swtpm_localca state may not be accessible so create our own.


etc/swtpm_setup.conf
```diff
@@ -0,0 +1,2 @@
+create_certs_tool=swtpm_localca
+create_certs_tool_config=etc/swtpm-localca.conf
```
kostyanf14, line 2: do we need create_certs_tool_options or this is optional?

etc/swtpm-localca.conf
```diff
@@ -0,0 +1,3 @@
+statedir=${AUTOHCK_STATEDIR}
+signingkey=${AUTOHCK_STATEDIR}/signkey.pem
+issuercert = ${AUTOHCK_STATEDIR}/issuercert.pem
```
kostyanf14, line 3: Do we need certserial or this is optional?

kostyanf14 (commented):
Looks good, just have questions about the config after compared with the original /etc files.

etc/swtpm_setup.conf
akihikodaki: It is necessary. `etc/swtpm-localca.conf` contains paths to files that must be relocated to the workspace.

akihikodaki: You mentioned `create_certs_tool_options`. I confirmed it works without the parameter.

etc/swtpm-localca.conf
akihikodaki: It is optional.

kostyanf14 (approved):

PR #337, opened 2024-04-05
Terminate CI if GitHub PR contains SKIP-HCK-CI

kostyanf14:
This allow to have build only PRs.

### DIFF
--- a/lib/auxiliary/github.rb
+++ b/lib/auxiliary/github.rb
@@ -39,6 +39,16 @@ module AutoHCK
       pr_object.state == 'closed'
     end
 
+    def pr_check_run(pr_object = find_pr)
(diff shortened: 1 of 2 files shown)


akihikodaki (approved):

lib/auxiliary/github.rb
```diff
@@ -39,6 +39,16 @@ def pr_closed?(pr_object = find_pr)
       pr_object.state == 'closed'
     end
 
+    def pr_check_run(pr_object = find_pr)
+      unless (pr_object['title'] =~ /SKIP-HCK-CI/).nil? && (pr_object['body'] =~ /SKIP-HCK-CI/).nil?
```
YanVugenfirer, line 43: Maybe use config file with dictionary?

Just asking if we would like in th future to have different behavior related to trigger file. Like don't trigger all tests , only specific driver 

Not a must - just to think about for the future

PR #348, opened 2024-04-17
Allow test duplication

kostyanf14:
This is experimental feature for dev testing.
Test results table can be broken.

Add option to have the same behavior if someone
use playlists with test duplication.


lib/engines/hcktest/playlist.rb
```diff
@@ -82,6 +82,17 @@ def sort_by_duration
       @tests.sort_by! { |test| test['duration'] }
     end
 
+    def intersect_select_tests(select_test_names)
+      select_test_names_counted = select_test_names.tally
+
+      @tests = @tests.filter_map do |test|
```
akihikodaki, line 88: Use `select!`.

lib/engines/hcktest/playlist.rb
```diff
@@ -82,6 +82,17 @@ def sort_by_duration
...
+        if select_test_names_counted.key?(test['name'])
+          test['run_count'] = select_test_names_counted[test['name']] if @project.options.test.allow_test_duplication
```
akihikodaki, line 90: Assign `1` to `run_count` when not allowing test duplication. It's better to have a consistent data layout whether allowing test duplication or not.

lib/engines/hcktest/tests.rb
```diff
@@ -285,8 +285,21 @@ def delete_old_remote(test_name)
...
     def all_tests_finished?
-      status_count('InQueue').zero? && current_test.nil?
+      # After the first Pass, HLK always reports a Pass as
+      # a status for this test even if it was queued again.
+      # Check that the result of the last queued test is not
+      # indicated as running.
+      # current_test can have a race condition when AutoHCK
+      # requests it earlier than the HLK update test execution state.
+      status_count('InQueue').zero? && current_test.nil? && !last_queued_test_running?
     end
```
akihikodaki, line 303: Shouldn't `current_test` be fixed instead of fixing `all_tests_finished?` as `current_test` is also used in other places?

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -426,10 +427,17 @@ def run
...
+          test_str = "run #{run_number}/#{run_count} #{test['name']} (#{test['id']})"
+          @logger.info("Adding to queue: #{test_str}")
+          queue_test(test, wait: true)
+          list_tests
+          handle_test_running
+
+          break if @project.run_terminated
+        end
 
         break if @project.run_terminated
```
akihikodaki, line 442: This line is no longer needed.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
kostyanf14: Why? We should interrupt 2 loops. Maybe `return if @project.run_terminated`?

akihikodaki: I was wrong. It's fine in the current form.

Jedoku (approved):

PR #353, opened 2024-04-25
Add cli option to enable debug mode in HLK-Setup-Scripts (install engine only)

Jedoku:


akihikodaki (approved):

lib/cli.rb
```diff
@@ -43,21 +43,21 @@ def create_parser(sub_parser)
...
         version_option(parser)
       end
 
-      def debug_option(parser)
-        parser.on('--debug', TrueClass,
-                  'Printing debug information') do |debug|
-          @debug = debug
+      def verbose_option(parser)
+        parser.on('--verbose', TrueClass,
+                  'Printing verbose information') do |verbose|
```
kostyanf14, line 59: Maybe it is better to `Enable verbose logging` this is more detailed.

PR #354, opened 2024-04-25
Add transfer_net common options

Jedoku:


lib/cli.rb
```diff
@@ -47,13 +47,23 @@ def define_options(parser)
...
+        client_transfer_net_option(parser)
         verbose_option(parser)
         config_option(parser)
         client_world_net_option(parser)
         id_option(parser)
         version_option(parser)
       end
 
+      def client_transfer_net_option(parser)
+        parser.on('--client_transfer_net', TrueClass,
```
kostyanf14, line 61: 1. This is not client specific. transfer_net enabled it attaches to all VMs.
2. We still need to manually change the config to specify `share_on_host_path`. As a variant, make `client_transfer_net` a string option that gets the path to a folder that will be shared. @akihikodaki What do you think?

kostyanf14 (changes requested):

lib/cli.rb
akihikodaki: > As a variant, make `client_transfer_net` a string option that gets the path to a folder that will be shared.

This sounds good, but it is better to be named `share_on_host_path` instead. Sharing a directory when a network is requested is odd. Attaching a network when sharing a directory is fine as a shared directory cleary depends on a network.

lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -156,7 +156,7 @@ def transfer_device_command(device_name, transfer_net, share_path, qemu_replacem
         net_base = "#{transfer_net}.0/24"
         smb_server = "#{transfer_net}.4"
 
-        netdev_options = "net=#{net_base},smb=#{path},smbserver=#{smb_server},restrict=on,ifname=@net_if_name@"
```
Jedoku, line 159: looks like ifname parameter not supporting in User mode networking

lib/cli.rb
```diff
@@ -127,6 +136,14 @@ def define_options(parser)
         reject_report_sections_option(parser)
         boot_device_option(parser)
         allow_test_duplication_option(parser)
+        share_on_host_net_option(parser)
+      end
+
+      def share_on_host_net_option(parser)
+        parser.on('--share-on-host-net <share_on_host_net>', String,
```
akihikodaki, line 143: It is a misnomer. We don't share a directory on the host network. We share a directory.

lib/cli.rb
```diff
@@ -127,6 +136,14 @@ def define_options(parser)
...
+                  'Use custom transfer network') do |share_on_host_net|
```
akihikodaki, line 144: This description also needs to be updated. The option is to share a directory. It creates a network for that, but it is just a side effect.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/network_manager.rb
kostyanf14: remove it

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -74,7 +74,8 @@ def platform_options
       OPT_NAMES.each do |name|
         options[name] = @platform[name] unless @platform[name].nil?
       end
-      options
+      share_host_directory = @project.options.common.share_host_directory || @project.options.test.share_host_directory
+      options.merge({ 'share_host_directory' => share_host_directory })
```
akihikodaki, line 78: This should not be added here as it is not a platform option (that can be defined with a JSON file in `lib/engines/hcktest/platforms` or with `platforms_defaults` in `lib/setupmanagers/qemuhck/qemu_machine.json`).

akihikodaki (changes requested):
Keep names consistent; the same thing should not be called two different names, `share_host_directory` and `share_on_host_path`.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -92,7 +98,7 @@ def client_vm_common_options
         'slirp' => @slirp,
         'iso_path' => @project.config['iso_path'],
         'client_world_net' => @project.options.common.client_world_net
-      }.merge(boot_device)
+      }.merge(boot_device).merge(share_on_host_path)
```
akihikodaki, line 101:
```suggestion
      }.merge(boot_device, share_on_host_path)
```

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -83,6 +83,12 @@ def boot_device
       { 'boot_device' => @project.options.test.boot_device }
     end
 
+    def share_on_host_path
+      share_on_host_path = @project.options.common.share_on_host_path ||
+                           @project.options.test.share_on_host_path
+      { 'share_on_host_path' => share_on_host_path || nil }
```
akihikodaki, line 89:
```suggestion
      { 'share_on_host_path' => share_on_host_path }
```

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -436,11 +436,12 @@ def process_device(device_info)
     end
 
     def process_optional_hck_network
-      return unless @config['transfer_net_enabled']
+      return if !@config['transfer_net_enabled'] && !@options['share_on_host_path']
```
akihikodaki, line 439: It is awkward to create a transfer network when `transfer_net_enabled` is false. What about removing it?

akihikodaki (changes requested):

lib/cli.rb
```diff
@@ -47,11 +47,20 @@ def define_options(parser)
...
         client_world_net_option(parser)
         id_option(parser)
         version_option(parser)
+        share_on_host_path_option(parser)
+      end
+
+      def share_on_host_path_option(parser)
+        parser.on('--share-on-host-path <path>', String,
+                  'Specify the directory to share on host machine') do |share_on_host_path|
+          @share_on_host_path = share_on_host_path
```
kostyanf14, line 62: Why do you define the same option in Common and Test options? Common is enough.

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: `transfer_net_enabled` is a config option. If for some reason we want to have a share for every run in the same folder.

kostyanf14 (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: What about removing the config option?

kostyanf14: ok, let's remove it for now

lib/cli.rb
```diff
@@ -25,7 +25,7 @@ def initialize
```
kostyanf14, line 25: `Signed-off-by: Jedoku <vitalii@daynix.com>` - Fix with real name

kostyanf14 (approved):
@akihikodaki Please review

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -436,11 +436,11 @@ def process_device(device_info)
     end
 
     def process_optional_hck_network
-      return unless @config['transfer_net_enabled']
+      return unless @options['share_on_host_path']
 
       dev = @nm.transfer_device_command(@config['transfer_net_device'],
                                         @config['share_on_host_net'],
-                                        @config['share_on_host_path'],
+                                        @options['share_on_host_path'],
```
akihikodaki, line 443: It's better to make it configurable with `@config` so that I can leave file sharing always enabled on my development environment, for example.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: What about your comment 2 week ago:
```
It is awkward to create a transfer network when transfer_net_enabled is false. What about removing it?
```

akihikodaki: I'm not for keeping `transfer_net_enabled` in `@config`, but I'm suggesting keeping `share_on_host_path` in `@config`.

akihikodaki: What I'm suggesting is:
```Ruby
      path = @options['share_on_host_path'] || @config['share_on_host_path']
      return unless path

      dev = @nm.transfer_device_command(@config['transfer_net_device'],
                                        @config['share_on_host_net'],
                                        path,
                                        full_replacement_list)
      @device_commands << dev
```

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -83,6 +83,12 @@ def boot_device
       { 'boot_device' => @project.options.test.boot_device }
     end
 
+    def share_on_host_path
+      return {} if @project.options.common.share_on_host_path.nil?
```
akihikodaki, line 87: This line should no longer be necessary.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -83,6 +83,10 @@ def boot_device
       { 'boot_device' => @project.options.test.boot_device }
     end
 
+    def share_on_host_path
+      { 'share_on_host_path' => @project.options.common.share_on_host_path }
+    end
+
```
akihikodaki, line 89: Why not just put it into the Hash literals in `studio_vm_options` and `client_vm_common_options` like other values?

akihikodaki (changes requested):

akihikodaki (approved):

PR #356, opened 2024-04-30
Misc changes related to SMB

akihikodaki:
QEMU binds smbd to lo so it needs to be up.

### DIFF
--- a/lib/setupmanagers/qemuhck/devices/tpm-tis.json
+++ b/lib/setupmanagers/qemuhck/devices/tpm-tis.json
@@ -8,7 +8,7 @@
   "define_variables": {
     "@tpm_localca_dir@": "@workspace@/swtpm_localca_@run_id@_@client_id@",
     "@tpm_state_dir@": "@workspace@/swtpm_state_@run_id@_@client_id@",
-    "@tpm_socket@": "/tmp/swtpm_@run_id@_@client_id@_sock"
+    "@tpm_socket@": "swtpm_@run_id@_@client_id@_sock"
   },
   "config_commands": [
     "mkdir -p @tpm_localca_dir@ @tpm_state_dir@",
--- a/lib/setupmanagers/qemuhck/devices/tpm-tis.json
+++ b/lib/setupmanagers/qemuhck/devices/tpm-tis.json
@@ -6,7 +6,6 @@
     "-device tpm-tis,tpmdev=tpm0"
   ],
   "define_variables": {
-    "@tpm_pid@": "@workspace@/swtpm_@run_id@_@client_id@.pid",
     "@tpm_localca_dir@": "@workspace@/swtpm_localca_@run_id@_@client_id@",
     "@tpm_state_dir@": "@workspace@/swtpm_state_@run_id@_@client_id@",
     "@tpm_socket@": "/tmp/swtpm_@run_id@_@client_id@_sock"
@@ -17,9 +16,6 @@
   ],
   "pre_start_commands": [
     "rm -f @tpm_socket@",
-    "swtpm socket --daemon --pid file=@tpm_pid@ --tpm2 --tpmstate dir=@tpm_state_dir@ --ctrl type=unixio,path=@tpm_socket@ --log file=@tpm_state_dir@/swtpm.log,level=20"
-  ],
-  "post_stop_commands": [
-    "kill $(cat @tpm_pid@) || true"
+    "@source@/bin/unix_spawn @tpm_socket@ swtpm socket --tpm2 --tpmstate dir=@tpm_state_dir@ --ctrl type=unixio,fd=3 --log file=@tpm_state_dir@/swtpm.log,level=20"
   ]
 }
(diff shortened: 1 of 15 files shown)


lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -159,15 +168,28 @@ def transfer_device_command(device_name, transfer_net, share_path, qemu_replacem
...
+        net_smb_private_shellescaped = net_smb_private.shellescape
+
+        # Mapping the current user to root will make smbd assume it can drop and
+        # re-add supplementary groups. However, re-adding supplementary groups
+        # do not work because they are typically not mapped in the current user
+        # namespace. Map the current user to nobody when running smbd so that
+        # smbd will run in the non-root mode.
+        net_smb_cmd = "unshare --map-user=nobody smbd -l #{net_smb_private_shellescaped} -s #{net_smb_private_shellescaped}/smb.conf"
+
+        netdev_options = ",net=#{net_base},guestfwd=:#{smb_server}:445-cmd:#{net_smb_cmd},restrict=on"
```
kostyanf14, line 181: Are you sure that `smb_server` IP is correct? When QEMU starts smb+slirp it assigns  `<IP>.4` by default.

kostyanf14: SMB uses 2 ports 445 and 139, better to share both. 
```SMB ports are generally port numbers 139 and 445. Port 139 is used by SMB dialects that communicate over NetBIOS.```

File: lib/setupmanagers/qemuhck/devices/tpm-tis.json [hunk 2]
```diff
@@ -13,7 +13,7 @@
   },
   "config_commands": [
     "mkdir -p @tpm_localca_dir@ @tpm_state_dir@",
-    "AUTOHCK_STATEDIR=@tpm_localca_dir@ swtpm_setup --tpm-state @tpm_state_dir@ --tpm2 --createek --create-ek-cert --create-platform-cert --lock-nvram --config etc/swtpm_setup.conf"
+    "env AUTOHCK_STATEDIR=@tpm_localca_dir@ swtpm_setup --tpm-state @tpm_state_dir@ --tpm2 --createek --create-ek-cert --create-platform-cert --lock-nvram --config etc/swtpm_setup.conf"
```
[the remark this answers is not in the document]
akihikodaki: Because we set an environment variable here and we can no longer rely on the shell to do that.

lib/setupmanagers/qemuhck/network_manager.rb
akihikodaki: We don't let activate QEMU built-in smb logic so it should be safe to the address.
NetBIOS is quite old and it is a common practice to disable it for security if I remember correctly. Probably it's not worth adding code for such an obsolete protocol.

kostyanf14: Ok, will test and update you

kostyanf14:
@akihikodaki Please fix RuboCop CI

lib/setupmanagers/qemuhck/network_manager.rb
```diff
@@ -156,15 +168,28 @@ def transfer_device_command(device_name, transfer_net, share_path, qemu_replacem
         net_base = "#{transfer_net}.0/24"
         smb_server = "#{transfer_net}.4"
 
-        netdev_options = "net=#{net_base},smb=#{path},smbserver=#{smb_server},restrict=on,ifname=@net_if_name@"
+        net_smb_private = replace_string_recursive('@workspace@/smb_@run_id@', qemu_replacement_list)
```
kostyanf14, line 171: In my setup, I have a long workspace path and during the testing of this PR smbd failed with the error `messaging_dgm_ref failed: File name too long`. smbd creates a socket in this directory and Linux limits socket path to 108 chars long, so need to make sure with this

akihikodaki: Isn't it better to relocate the workspace? We already have other socket files in workspaces.

akihikodaki: > > We already have other socket files in workspaces.
> 
> which one? viofs socket moved to /tmp we have the same issue.

I missed it's in `/tmp`. I'll relocate smb sockets to `/tmp` too.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -264,21 +268,22 @@ def read_machine
...
+      @fw = fws[@fw_name]
 
-      unless res
+      unless @fw
         @logger.fatal("#{@fw_name} does not exist")
         raise(InvalidConfigFile, "#{@fw_name} does not exist")
       end
 
-      return res unless res['nvram']
+      @machine_options << 'pflash0=pflash_code' if @fw['binary']
```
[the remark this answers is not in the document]
akihikodaki: `fw_cmd` is called each time `run` is called so adding `machine_options` in `fw_cmd` will result in duplicate `machine_options` if `run` gets called multiple times.

kostyanf14 (approved):

PR #359, opened 2024-04-30
Change the meaning of manual mode 

kostyanf14:


Jedoku (approved):

lib/engines/hcktest/hcktest.rb
```diff
@@ -226,29 +226,44 @@ def tag
...
+      @project.logger.info('Manual exit. AutoHCK will continue.')
+    end
+
     def auto_run
       ResourceScope.open do |scope|
         run_and_configure_setup scope
         client = @client1
         client.run_tests
+
+        wait_manual
```
akihikodaki, line 254: It is awkward to call `wait_manual` from `auto_run`.

lib/engines/hcktest/hcktest.rb
```diff
@@ -226,29 +226,44 @@ def tag
...
 
+    def wait_manual
+      return unless @project.options.test.manual
+
+      @project.logger.info('AutoHCK started in manual mode. Waiting for manual exit.')
+      loop do
+        @project.logger.info("Type 'exit' and press ENTER to exit AutoHCK")
+
+        break if $stdin.gets&.chomp == 'exit'
+      end
```
akihikodaki, line 244: What about invoking irb? c.f., https://docs.ruby-lang.org/en/3.3/Binding.html#method-i-irb

akihikodaki (changes requested):
Perhaps it may be a good time to rename `dump_only` to `dump`. `dump_only` sounds as if AutoHCK usually dumps commands and run them and the option is to perform only command dumping, but in reality, command dumps are generated only when the option is specified.

lib/engines/hcktest/hcktest.rb
kostyanf14: renamed

lib/engines/hcktest/hcktest.rb
kostyanf14: 1. RuboCop does not like this
2. I don't like this allows users to access a lot of internal states.

akihikodaki: 1. RuboCop is confused and thinks you left code for debugging. Just ignore it.
2. You can create a `Binding` at a different scope to limit the state visibility. It is also possible to use `Ractor` or even create another Ruby process and perform IPC for strong isolation, but I don't think it's worthwhile.

lib/engines/hcktest/hcktest.rb
```diff
@@ -226,29 +223,59 @@ def tag
...
+        @project.logger.info("Type 'exit' and press ENTER to exit manul mode")
+
+        break if $stdin.gets&.chomp == 'exit'
+      end
+      @project.logger.info('Manual exit. AutoHCK will continue.')
+    end
+
+    def prepare_tests
+      client = @clients.values[0]
+      support = @clients.values[1]
```
akihikodaki, line 245:
```suggestion
      client, support = @clients.values
```

akihikodaki (changes requested):

akihikodaki (approved):

akihikodaki:
For `Binding#irb`, opened: https://github.com/sorbet/sorbet/pull/7889

PR #366, opened 2024-05-13
Add useful info to log

Jedoku:
https://issues.redhat.com/browse/RHELMISC-4835

current host.txt looks like: 

image

here is logs.txt output
image

### DIFF
--- a/lib/engines/hcktest/tests.rb
+++ b/lib/engines/hcktest/tests.rb
@@ -249,8 +249,18 @@ module AutoHCK
       @tests.reduce('') do |sum, test|
         extra_info = @tests_extra.dig(test['id'], 'dump') ? '(with Minidump)' : ''
         status = @tests_extra.dig(test['id'], 'status') || test['status']
-        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}\n"
+
+        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}]#{extra_info}#{format_times(test)}\n"
       end
+    end
+
+    def format_times(test)
+      queued_at = @tests_extra.dig(test['id'], 'queued_at')
+      queued_at_str = queued_at ? " [Queued time: #{queued_at}]" : ''
+      started_at = @tests_extra.dig(test['id'], 'started_at')
+      started_at_str = started_at ? " [Started time: #{started_at}]" : ''
+
+      "#{queued_at_str}#{started_at_str}"
     end
 
     def update_summary_results_log
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -66,7 +66,7 @@ module AutoHCK
         @run_opts = run_opts
         @keep_alive = run_opts[:keep_alive]
         @delete_snapshot = run_opts[:create_snapshot]
-        @machine.dump_config
+        @logger.info(@machine.dump_config)
         @qemu_thread = Thread.new { run_vm }
         scope << self
       end
@@ -581,7 +581,7 @@ module AutoHCK
       #  S3 enabled..................${ENABLE_S3}
       #  S4 enabled..................${ENABLE_S4}
       #  Snapshot mode.............. ${SNAPSHOT}
-      @logger.info(full_replacement_map.replace(config.join("\n")))
+      full_replacement_map.replace(config.join("\n"))
     end
 
     def create_run_script(file_name, file_content)
@@ -660,7 +660,7 @@ module AutoHCK
     end
 
     def dump_commands
-      dump_config
+      @logger.info(dump_config)
 
       file_name = "#{@workspace_path}/#{@run_name}_manual.sh"
       content = <<~BASH
--- a/lib/setupmanagers/qemuhck/qemuhck.rb
+++ b/lib/setupmanagers/qemuhck/qemuhck.rb
@@ -14,6 +14,7 @@ module AutoHCK
 
     attr_reader :kit, :project
 
+    QEMUHCK_INFO_LOG_FILE = 'qemuhck.txt'
     OPT_NAMES = %w[viommu_state s3_state s4_state enlightenments_state vhost_state machine_type fw_type cpu
                    pluggable_memory_gb].freeze
 
@@ -24,6 +25,7 @@ module AutoHCK
       @clients_vm = {}
       initialize_studio_vm
       initialize_clients_vm
+      create_qemuhck_log_file
     end
 
     def initialize_project(project)
@@ -105,6 +107,40 @@ module AutoHCK
           'memory_gb' => v['memory_gb']
         }.merge(client_vm_common_options))
         @clients_vm[v['name']] = QemuMachine.new(vm_options)
+      end
+    end
+
+    def create_qemuhck_log_file
+      qemuhck_log = "#{@workspace_path}/#{QEMUHCK_INFO_LOG_FILE}"
+      File.write(qemuhck_log, collect_environment_info)
+      @project.result_uploader.upload_file(qemuhck_log, QEMUHCK_INFO_LOG_FILE)
+    end
+
+    def collect_environment_info
+      logs = String.new
+      append_host_info(logs)
+      append_vms_info(logs)
+      logs.freeze
+    end
+
+    def append_host_info(logs)
+      logs << <<~HOST_INFO
+        QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}
+        System information: #{`uname -a`.strip}
+
+      HOST_INFO
+    end
+
+    def append_vms_info(logs)
+      logs << <<~STUDIO_INFO
+        Studio Properties:
+          #{@studio_vm.dump_config}
+      STUDIO_INFO
+      @platform['clients'].each_with_index do |(_k, v), i|
+        logs << <<~CLIENT_INFO
+          Client #{i + 1} Properties:
+            #{@clients_vm[v['name']].dump_config}
+        CLIENT_INFO
       end
     end
 


lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +111,33 @@ def initialize_clients_vm
...
+    def compose_host_info
+      logs = []
+      logs << "Hostname: #{`hostname`.strip}"
+      logs << "QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}"
+      logs << "Linux OS info: #{parse_os_release}"
+      logs << "Linux Kernel version: #{`uname -a`.strip}"
+      logs << "\nVMs Info:"
+
+      @vm_info.each do |vm|
+        logs << "VM Name: #{vm[:name]}, VNC Port: #{vm[:vnc_port]}"
```
YanVugenfirer, line 136: Do we also need an HMP\QMP port? It can be useful. Do we start QMP or HMP monitor?

lib/engines/hcktest/tests.rb
```diff
@@ -249,10 +249,22 @@ def summary_results_log
...
         status = @tests_extra.dig(test['id'], 'status') || test['status']
-        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}\n"
+        queued_at, started_at = format_times(test)
+
+        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}#{queued_at}#{started_at}\n"
       end
     end
 
+    def format_times(test)
+      queued_at = @tests_extra.dig(test['id'], 'queued_at')&.strftime('%Y-%m-%d %H:%M:%S')
```
kostyanf14, line 259: After `strftime` you don't have a timezone in the string, so it is not clear what exactly the time is when CI has a different timezone from you. Don't format time as we don't do this in any user-visible reports.

lib/setupmanagers/qemuhck/qemuhck.rb
akihikodaki: We start QMP, but it listens on a UNIX domain socket.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +111,33 @@ def initialize_clients_vm
       end
     end
 
+    def write_host_info
+      collect_vms_info
```
akihikodaki, line 115: It's odd to collect "VM" information when writing host information.
Why will you want to log VM information anyway? AutoHCK already logs the whole QEMU command line.

lib/setupmanagers/qemuhck/qemuhck.rb
kostyanf14: Yes, we have HMP and QMP. Let's add it.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +111,33 @@ def initialize_clients_vm
...
+      @vm_info = []
+      @vm_info << { name: 'studio', vnc_port: @studio_vm.vnc_port }
+      @platform['clients'].each do |(_k, v)|
+        @vm_info << { name: v['name'], vnc_port: @clients_vm[v['name']].vnc_port }
+      end
+    end
+
+    def compose_host_info
+      logs = []
+      logs << "Hostname: #{`hostname`.strip}"
```
akihikodaki, line 129: `uname -a` already contains the host name so this is redundant.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +111,33 @@ def initialize_clients_vm
...
+      logs << "QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}"
+      logs << "Linux OS info: #{parse_os_release}"
+      logs << "Linux Kernel version: #{`uname -a`.strip}"
+      logs << "\nVMs Info:"
```
kostyanf14, line 133: Why do you need `\n` there?

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +111,33 @@ def initialize_clients_vm
...
+      @platform['clients'].each do |(_k, v)|
+        @vm_info << { name: v['name'], vnc_port: @clients_vm[v['name']].vnc_port }
+      end
+    end
+
+    def compose_host_info
+      logs = []
+      logs << "Hostname: #{`hostname`.strip}"
+      logs << "QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}"
+      logs << "Linux OS info: #{parse_os_release}"
```
akihikodaki, line 131: I doubt the utility of printing `/etc/os-release`. It is usually obvious what distribution is in use when `uname -a` is printed.

kostyanf14 (changes requested):
There is `host.txt` file creation but I didn't find uploading of it.

lib/setupmanagers/qemuhck/qemuhck.rb
kostyanf14: @akihikodaki We still have HMP via telnet port.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemuhck.rb
kostyanf14: `AutoHCK.log` are uploaded only at the end, we want to have this info during the CI run to easily know how to connect to the running env.

lib/engines/hcktest/tests.rb
```diff
@@ -249,10 +249,22 @@ def summary_results_log
       @tests.reduce('') do |sum, test|
```
kostyanf14, line 249: `Signed-off-by: Jedoku <vitalii@daynix.com>` - Fix with real name

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -545,6 +545,7 @@ def find_world_ip
     def dump_config
       config = [
         'Setup configuration:',
+        "   VM name .................... CL#{@options['client_id']}",
```
kostyanf14, line 548:
```Studio Properties:
Setup configuration:
VM name .................... CL0
```

As we have a `Studio Properties` we can skip `VM name` line, also `CL0` is not so clear.

lib/engines/hcktest/tests.rb
```diff
@@ -249,10 +249,22 @@ def summary_results_log
...
         status = @tests_extra.dig(test['id'], 'status') || test['status']
-        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}\n"
+        queued_at, started_at = format_times(test)
+
+        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}#{queued_at}#{started_at}\n"
       end
     end
 
+    def format_times(test)
+      queued_at = @tests_extra.dig(test['id'], 'queued_at')&.strftime('%Y-%m-%d %H:%M:%S %Z')
```
akihikodaki, line 259: Probably you don't need `strftime` call at all.
`" [Queued time: #{Time.now}]"` evaluates to: `" [Queued time: 2024-05-14 12:19:27 +0900]"`

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,35 @@ def initialize_clients_vm
       end
     end
 
+    def write_host_info
+      host_log = "#{@workspace_path}/#{HOST_INFO_LOG_FILE}"
```
akihikodaki, line 114: These should also renamed.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,33 @@ def initialize_clients_vm
...
+
+    def collect_environment_info
+      logs = []
+      compose_host_info(logs)
+      compose_vms_info(logs)
+      logs
+    end
+
+    def compose_host_info(logs)
+      logs << "QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}"
```
akihikodaki, line 127: Use heredoc.

lib/engines/hcktest/tests.rb
```diff
@@ -249,10 +249,21 @@ def summary_results_log
       @tests.reduce('') do |sum, test|
         extra_info = @tests_extra.dig(test['id'], 'dump') ? '(with Minidump)' : ''
         status = @tests_extra.dig(test['id'], 'status') || test['status']
-        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}\n"
+        queued_at, started_at = format_times(test)
+
+        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}#{queued_at}#{started_at}\n"
```
akihikodaki, line 254:
```suggestion
        sum + "#{status}: #{test['name']} [#{test['estimatedruntime']}] #{extra_info}#{format_times(test)}\n"
```

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,33 @@ def initialize_clients_vm
...
     end
 
+    def write_host_file
+      host_log = "#{@workspace_path}/#{HOST_INFO_LOG_FILE}"
+      File.write(host_log, collect_environment_info.join("\n"))
+      @project.result_uploader.upload_file(host_log, HOST_INFO_LOG_FILE)
+    end
+
+    def collect_environment_info
+      logs = []
```
akihikodaki, line 120: Use a mutable `String` instead of an `Array`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/devices/iommu.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "iommu",
   "command_line": [
-    "-device intel-iommu,intremap=on,aw-bits=48,caching-mode=on,device-iotlb=on"
+    "-device amd-iommu"
```
kostyanf14, line 4: Not for this PR))

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -101,8 +101,8 @@ def check_fails_too_quickly(status)
         end
       end
 
-      def run_vm
-        @machine.dump_config
+      def run_vm # rubocop:disable Metrics/AbcSize
```
kostyanf14, line 104: Rebase to current master. This method updated. 
If the AbcSize metric still will be big, update the Rubocop config according to the Slack discussion.

kostyanf14 (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -212,6 +212,7 @@ def initialize(options)
       @devices_list.flatten!
       @devices_list.compact!
       load_devices
+      @image_path = @sm.boot_device_command(option_config('boot_device'), @run_opts, full_replacement_map).last
```
kostyanf14, line 215: boot_device_command calls create_boot_snapshot, so after you change QemuHCK will generate a snapshot 2 times. This looks bad and can cause unexpected behavior in the future like data loss or vice versa we will have a snapshot but expect that it does not exist.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -212,6 +212,7 @@ def initialize(options)
       @devices_list.flatten!
       @devices_list.compact!
       load_devices
+      @image_path = @sm.boot_image_path
```
akihikodaki, line 215: This is different from what the VM actually uses if `create_snapshot`.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,40 @@ def initialize_clients_vm
...
+      host_log = "#{@workspace_path}/#{HOST_INFO_LOG_FILE}"
+      File.write(host_log, collect_environment_info)
+      @project.result_uploader.upload_file(host_log, HOST_INFO_LOG_FILE)
+    end
+
+    def collect_environment_info
+      logs = String.new
+      compose_host_info(logs)
+      compose_vms_info(logs)
+      logs
```
akihikodaki, line 123: Freeze it before returning.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,40 @@ def initialize_clients_vm
...
+    end
+
+    def compose_host_info(logs)
```
akihikodaki, line 126: Let's name it like `append_host_info` to clarify it mutates `logs`.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,40 @@ def initialize_clients_vm
...
+        QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}
+        System information: #{`uname -a`.strip}
+      HOST_INFO
+    end
+
+    def compose_vms_info(logs)
+      logs << <<~STUDIO_INFO
+        Studio Properties:
+          #{@studio_vm.dump_config}
+
```
akihikodaki, line 137: For consistency do either:
- add a blank line after the host information.
- Remove the blank line after the studio properties.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -108,6 +110,40 @@ def initialize_clients_vm
...
+    def append_host_info(logs)
+      logs << <<~HOST_INFO
+        QEMU version: #{`qemu-system-x86_64 --version`.lines.first.strip}
+        System information: #{`uname -a`.strip}
+
+      HOST_INFO
+    end
+
+    def compose_vms_info(logs)
```
akihikodaki, line 134: This needs to be renamed like `compose_vms_info` too.

akihikodaki (changes requested):

kostyanf14 (approved):

akihikodaki (changes requested):
https://github.com/HCK-CI/AutoHCK/pull/366/files#diff-3e6726d4233abd2e11fcb42d3d3cf38568b0b02510d78573e318f1add3fc5bd7 and https://github.com/HCK-CI/AutoHCK/pull/366/files#diff-0e4513135adb7f5e09f7b6318253ebe5071c3e59220a434a1725483fc1ebb2e8 still need to be addressed.

lib/setupmanagers/qemuhck/qemuhck.rb
akihikodaki: `@image_path` can be simply omitted from `dump_config`. I think that information is not more useful than other omitted information.

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: `@image_path` can be simply omitted from `dump_config`. I think that information is not more useful than other omitted information.

lib/setupmanagers/qemuhck/qemuhck.rb
akihikodaki: `host.txt` can be renamed to `qemuhck.txt` since it's specific to QemuHCK.

Jedoku: The variable was updated to `@sm.boot_snapshot_path`, which is precisely the path we require.

lib/setupmanagers/qemuhck/qemu_machine.rb
Jedoku: The variable was updated to @sm.boot_snapshot_path, which is precisely the path we require.

lib/setupmanagers/qemuhck/qemuhck.rb
Jedoku: The function is designed to gather host information, and we also include additional details about QEMU VMs since that data is readily available. Initially, it should capture several ports.

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: > The variable was updated to @sm.boot_snapshot_path, which is precisely the path we require.

It makes the printed path wrong when `create_snapshot` is false.

lib/setupmanagers/qemuhck/qemuhck.rb
akihikodaki: > The function is designed to gather host information, and we also include additional details about QEMU VMs since that data is readily available. Initially, it should capture several ports.

One of the original use cases is to observe ports VMs expose in my understanding. In this sense, this feature is not to expose the host information, but to expose information necessary to manage QEMU, and that includes VMs. That's why it is misleading to call it "host information".

akihikodaki (approved):

kostyanf14 (approved):

PR #369, opened 2024-05-15
CI: Use reusing workflow for Setup Ruby env

kostyanf14:


akihikodaki (changes requested):
- `shell: bash` looks verbose. Can't you omit them?
- Some strings are quoted with double quotes, single quotes, and the others are not quoted at all. We can always omit quotes.
- `sqlite-devel` should no longer be necessary. It's a bit out of scope but it may be good time to remove it.

kostyanf14:
>shell: bash looks verbose. Can't you omit them?

No. Mandatory for action

>    Some strings are quoted with double quotes, single quotes, and the others are not quoted at all. We can always omit quotes.
>    sqlite-devel should no longer be necessary. It's a bit out of scope but it may be good time to remove it.

Will fix

akihikodaki:
> > shell: bash looks verbose. Can't you omit them?
> 
> No. Mandatory for action

It does not seem so as far as I read the documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

I forgot one thing to note: why not check out the code and name it "Set up AutoHCK" since it already assumes there is a checkout to run Bundler?

kostyanf14:
>It does not seem so as far as I read the documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

Very strange. CI failed without shell.

> why not check out the code and name it "Set up AutoHCK" since it already assumes there is a checkout to run Bundler?

to use local action/workflow (from the same repo) you must run checkout first
rename ok

akihikodaki:
> > It does not seem so as far as I read the documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
> 
> Very strange. CI failed without shell.
> 
> > why not check out the code and name it "Set up AutoHCK" since it already assumes there is a checkout to run Bundler?
> 
> to use local action/workflow (from the same repo) you must run checkout first rename ok

I meant to check out the code in the composite action instead of checking out in each jobs.

akihikodaki:
> > It does not seem so as far as I read the documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
> 
> Very strange. CI failed without shell.

I found: https://github.com/orgs/community/discussions/18597
Unfortunately, it seems to be required for composite actions.

akihikodaki (approved):

PR #376, opened 2024-06-09
QemuHCK: Automatically enter a namespace

akihikodaki:
I'm working on improving `bin/ns`, but it seems to take a bit longer so let me submit dependent changes early here.


bin/triggers_check
```diff
@@ -4,7 +4,7 @@
 require 'bundler/setup'
 require 'dotenv/load'
 require 'optparse'
```
kostyanf14: `require 'optparse'` can be removed too?

lib/cli.rb
```diff
@@ -233,9 +237,9 @@ def define_options(parser)
     end
 
     def parse(args)
-      @parser.order!(args)
-      @mode = args.shift
-      @sub_parser[@mode]&.order!(args) unless @mode.nil?
+      left = @parser.order(args)
+      @mode = left.shift
+      @sub_parser[@mode]&.order!(left) unless @mode.nil?
```
kostyanf14, line 242: why do we need this? behaviour is the same

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -32,17 +32,12 @@ def self.tag(options)
...
 
       @logger.info("Loading platform: #{platform_name}")
-      unless File.exist?(platform_json)
-        @logger.fatal("#{platform_name} does not exist")
-        raise(InvalidConfigFile, "#{platform_name} does not exist")
-      end
-
-      Json.read_json(platform_json, @logger)
+      Json.read_json(platform_json, logger)
     end
```
kostyanf14, line 41: looks like it should be in commit  `Do not check file existence before calling Json.read_json`

bin/triggers_check
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+Dir.chdir(File.dirname(__dir__))
+
```
kostyanf14: Why don't you place this in commit `Add AutoHCK.run` ?

lib/cli.rb
akihikodaki: `@parser.order!` modifies `args`, which is problematic because we use `args` again when entering a namespace.

bin/triggers_check
akihikodaki: I left it because it's used in this file.

lib/engines/hckinstall/hckinstall.rb
akihikodaki: This method has two changes. The file existence check gets removed first, and then it gets converted into a class method. This line is changed due to the latter conversion.

bin/triggers_check
akihikodaki: Removed with commit 49e76f06c369f0e37aced501a1f3d8d7c01be184.

bin/slirp
```diff
@@ -0,0 +1,73 @@
...
+
+module AutoHCK
+  # rubocop:disable Metrics/BlockLength
+  run do
+    # rubocop:enable Metrics/BlockLength
+
+    require 'optparse'
+    require 'socket'
+
+    t = nil
```
Jedoku, line 16: Can we name it better?

akihikodaki: I named it after the command line option. The meaning should be obvious if you know the option.

akihikodaki: Renamed the variable to `workspace_path` to match with the rest of the codebase. The command line option itself has also renamed to match with `bin/auto_hck`'s `w` option.

akihikodaki:
Rebased the series and fixed the VNC connection.

lib/project.rb
```diff
@@ -170,6 +173,7 @@ def init_workspace
         @logger.warn('Workspace path already exists')
       end
       @logger.info("Workspace path: #{@workspace_path}")
+      @setup_manager_type&.enter @workspace_path
```
kostyanf14, line 174: Why we don't need to enter workspace_path if it is specified? 

I am thinking about the next task: `Add ability to bring up previous session (for example to get dump from)` 

I expect that  auto_hck will load/initialize the state of every component when  workspace_path is provided.

Maybe we should have 2 different options for this one for internal use and the other for public usage?

bin/ns
```diff
@@ -4,91 +4,29 @@
...
+    parser.version = VERSION
+    parser.on('-w <path>', '--workspace <path>', String, 'Workspace path') { workspace_path = _1 }
+    argv = parser.order(ARGV)
+
+    if workspace_path.nil?
+      Dir.chdir Config.read['workspace_path']
+      dirs = Dir['*/*/*/']
+      abort 'no workspace found' if dirs.empty?
+
+      workspace_path = dirs.max_by { File.mtime(_1) }
```
kostyanf14, line 25: This can be broken when several instances of AutoHCK run in parallel. I think we should fail without a workspace.

lib/setupmanagers/qemuhck/qemuhck.rb
```diff
@@ -4,6 +4,9 @@
 module AutoHCK
   # QemuHCK class
   class QemuHCK
+    extend AutoloadExtension
+    autoload_relative :Ns, 'ns'
+
```
kostyanf14, line 9: why not in `all.rb`?

lib/setupmanagers/qemuhck/ns.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module AutoHCK
+  class QemuHCK
+    module Ns
+      extend AutoloadExtension
+      autoload_relative :Nsd, 'ns/nsd'
+      autoload_relative :PidFile, 'ns/pid_file'
```
kostyanf14, line 8: why not in `all.rb`?

kostyanf14 (changes requested):

bin/ns
akihikodaki: It is a convenient fallback for development. You usually don't manually invoke `bin/ns`.

lib/setupmanagers/qemuhck/qemuhck.rb
akihikodaki: Because `Ns` is in `AutoHCK::QemuHCK`.

lib/project.rb
akihikodaki: > Maybe we should have 2 different options for this one for internal use and the other for public usage?

Perhaps so, but it is out of scope of this change.

kostyanf14 (approved):

PR #390, opened 2024-06-24
Fix custom user text playlist does not work

kostyanf14:


akihikodaki (commented):
What if `select_test_names` is not specified?

kostyanf14:
> What if `select_test_names` is not specified?

we have `return unless select_test_names` in `def custom_select_test_names`

akihikodaki:
> > What if `select_test_names` is not specified?
> 
> we have `return unless select_test_names` in `def custom_select_test_names`

Won't we need to set `run_count` in that case too?

kostyanf14:
> Won't we need to set `run_count` in that case too?

We set `run_count` in list_tests for default value 1

lib/engines/hcktest/playlist.rb
```diff
@@ -90,14 +90,16 @@ def sort_by_duration
...
 
       @tests.select! do |test|
         next unless select_test_names_counted.key?(test['name'])
 
-        test['run_count'] = select_test_names_counted[test['name']]
+        test['run_count'] = if @project.options.test.allow_test_duplication
+                              select_test_names_counted[test['name']]
+                            else
+                              1
+                            end
```
akihikodaki, line 102:
```suggestion
        if @project.options.test.allow_test_duplication
          test['run_count'] = select_test_names_counted[test['name']]
        end
        
        true
```

akihikodaki (changes requested):

akihikodaki (approved):

PR #413, opened 2024-07-10
RFC: Boot order redesign 

kostyanf14:


lib/setupmanagers/qemuhck/storage_manager.rb
```diff
@@ -83,18 +84,32 @@ def boot_device_command(device_name, run_opts, qemu_replacement_map)
...
+
+        run_opts[:attach_iso_list]&.map do |iso|
+          iso_path = Pathname.new(@iso_path).join(iso)
+          boot_index += 1
+
+          "-device ide-cd,drive=iso_drive_#{boot_index},bus=ide.#{boot_index},bootindex=#{boot_index} " \
+            "-drive file=#{iso_path},if=none,media=cdrom,readonly=on,id=iso_drive_#{boot_index}"
+        end
+      end
+
```
akihikodaki, line 105: Why would we want `bootindex` for ISOs?

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -504,7 +507,7 @@ def base_cmd
       [
         '@qemu_bin@ -enable-kvm -machine @machine_options@ ',
         '-m @memory@,maxmem=@max_memory@ -smp @cpu_count@,cores=@cpu_count@ ',
-        '-cpu @cpu_options@ -boot order=cd,menu=on ',
+        '-cpu @cpu_options@ -boot menu=on,splash-time=10000 ',
```
akihikodaki, line 510: Please add a comment to describe the situation where `splash-time` is needed.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/storage_manager.rb
akihikodaki: It's better to specify `bootindex` only for the install disk though I don't require to make a change for that.

kostyanf14: Yes, this is a better solution,  but storage_manager and qemuhck do not know what image is really bootable.

akihikodaki: A simple approach is to introduce a new option like `attach_installer_iso`.

kostyanf14: Will move to separate task

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: done

akihikodaki (approved):

PR #419, opened 2024-07-18
Add ability to run specific tests with Secure Boot

Jedoku:


akihikodaki (changes requested):
Don't add `stop`. Instead, open a `ResourceScope` and run clients in the scope for each test group.

lib/engines/hcktest/hcktest.rb
```diff
@@ -151,21 +151,28 @@ def configure_clients
       end
     end
 
+    def reconfigure_clients
```
akihikodaki, line 154: This name does not represent what it actually does. "Reconfiguring" sounds like it is a special operation that corresponds to `configure_clients` but should be used for a second time. But it does more than configure (it also synchronizes). And such a sequence of configuring and synchronizing is also valid for the first run.

lib/engines/hcktest/hcktest.rb
```diff
@@ -151,21 +151,28 @@ def configure_clients
...
+      configure_clients
+      @clients.each_value(&:synchronize)
+    end
+
     def configure_setup_and_synchronize
+      return reconfigure_clients unless @studio.tools.nil?
+
       @studio.configure(@platform['clients'])
       configure_clients
       @clients.each_value(&:synchronize)
```
akihikodaki, line 164: Call `reconfigure_clients` here (and `reconfigure_clients` should be renamed as it is awkward to call like that when it is used for the first run).

lib/engines/hcktest/hcktest.rb
```diff
@@ -151,21 +151,28 @@ def configure_clients
...
+    def run_and_configure_clients(scope)
       retries = 0
       begin
         scope.transaction do |tmp_scope|
-          run_studio tmp_scope
-          sleep 5 until @studio.up?
-          run_clients tmp_scope, keep_alive: true
+          run_opts = { keep_alive: true }
+          run_opts.merge!(@client_run_opts) if @client_run_opts
+          run_clients tmp_scope, run_opts
```
akihikodaki, line 175:
```suggestion
          run_clients tmp_scope, keep_alive: true, **@client_run_opts
```

lib/engines/hcktest/hcktest.rb
```diff
@@ -151,21 +151,28 @@ def configure_clients
...
+
       @studio.configure(@platform['clients'])
       configure_clients
       @clients.each_value(&:synchronize)
       @studio.keep_snapshot
       @clients.each_value(&:keep_snapshot)
     end
 
-    def run_and_configure_setup(scope)
+    def run_and_configure_clients(scope)
```
akihikodaki, line 169: Drop `setup` everywhere if you are dropping it here.

lib/engines/hcktest/hcktest.rb
akihikodaki: I also want you to minimize the use of instance variables; pass `@client_run_opts` as an argument of this method instead.

lib/engines/hcktest/hcktest.rb
```diff
@@ -225,22 +232,69 @@ def prepare_tests
...
+
+        prepare_environment(group)
+        ResourceScope.open do |scope|
+          run_and_configure_clients scope
+
+          @tests.run(tests)
+          break if @project.run_terminated
+        end
+      end
+    end
```
akihikodaki, line 265: This is a bit strange division of concerns. There are two things in `run_tests`; running "normal" tests and running the other tests. However, while the first is abstracted away, the second isn't.

We can extract the second away by creating another method. Not being "normal" is "abnormal" but that sounds bad in English. Perhaps you may name it e.g., `run_tests_without_config` and `run_tests_with_config` to represent what it does more precisely; it is verbose but I think that's fine since it's only called from one place anyway.

lib/engines/hcktest/hcktest.rb
```diff
@@ -225,22 +232,69 @@ def prepare_tests
...
+        sleep 5 until @studio.up?
 
-        @tests.run
+        run_tests
 
         pause_run if @project.options.test.manual
 
         @tests.create_project_package
       end
     end
```
akihikodaki, line 249: What should be done for `dump_run`?

akihikodaki (changes requested):

lib/engines/hcktest/hcktest.rb
```diff
@@ -225,19 +226,64 @@ def prepare_tests
...
+      grouped_tests = { normal: [], secure: [] }
+
+      @config.tests_config.each do |test_group|
+        selected_tests = @test_list.select { |test| test_group.tests.include?(test['name']) }
+        grouped_tests[:secure] += selected_tests if test_group.secure
+      end
+
+      grouped_tests[:normal] += (@test_list - grouped_tests.values.flatten)
+      grouped_tests
+    end
```
akihikodaki, line 280: Let's split this into a method that returns tests with config and another that returns tests without config.

lib/engines/hcktest/hcktest.rb
```diff
@@ -143,29 +143,30 @@ def run_clients(scope, run_opts = {})
...
       @studio.configure(@platform['clients'])
-      configure_clients
-      @clients.each_value(&:synchronize)
+      configure_and_synchronize_clients
       @studio.keep_snapshot
       @clients.each_value(&:keep_snapshot)
     end
 
-    def run_and_configure_setup(scope)
+    def run_clients_and_configure_setup(scope, client_run_opts = {})
```
akihikodaki, line 165: Simpler:

```suggestion
    def run_clients_and_configure_setup(scope, **opts)
```

We already say `run_clients` so we don't need to say that in this method. Also let's use `**` trick.

lib/engines/hcktest/hcktest.rb
```diff
@@ -225,19 +226,64 @@ def prepare_tests
...
+      end
+    end
 
-        @tests.create_project_package
+    def run_tests_with_config
+      group_tests.each do |group, tests|
+        next if tests.empty? || group == :normal
+
+        ResourceScope.open do |scope|
+          run_clients_and_configure_setup(scope, client_run_options(group))
```
akihikodaki, line 260: I guess you can just do:

```suggestion
          run_clients_and_configure_setup(scope, group => true)
```

lib/engines/hcktest/hcktest.rb
```diff
@@ -225,19 +226,64 @@ def prepare_tests
...
+
+          @tests.run(tests)
+
+          pause_run if @project.options.test.manual
+          break if @project.run_terminated
```
akihikodaki, line 265: Why do we need this line though we didn't have one previously and don't have one even now for the tests without config?

akihikodaki (changes requested):
I found the logging code is inconsistent. For the tests without config, it logs after configuring the clients. For tests with configs, it logs before configuring clients.

Let's also not introduce a new term "environment" but an existing word "clients".

lib/setupmanagers/hckclient.rb
```diff
@@ -134,6 +149,8 @@ def initialize_client_wait
     end
 
     def return_when_client_up
+      move_machine_to_default_pool if machine_in_working_pool?
```
kostyanf14, line 152: Why do we need this? Let's just skip recognize_client_wait and initialize_client_wait if client initialized and not in Default Pool.

lib/engines/hcktest/hcktest.rb
```diff
@@ -143,29 +143,30 @@ def run_clients(scope, run_opts = {})
...
 
-    def run_and_configure_setup(scope)
+    def run_clients_and_configure_setup(scope, client_run_opts = {})
       retries = 0
       begin
         scope.transaction do |tmp_scope|
-          run_studio tmp_scope
-          sleep 5 until @studio.up?
-          run_clients tmp_scope, keep_alive: true
+          run_clients tmp_scope, keep_alive: true, **client_run_opts
```
kostyanf14, line 169: I see a problem with the current implementation of HCKClient recreation. When we create a new HCKClient it will call setup_manager to create a snapshot again. In this case, we lose information from the previous run (can capture files from a real VM if needed). 

There are 2 ways:
- add a new option to reuse existing snapshot
- add a new option to rename snapshots according to the test group

@akihikodaki What do you think?

kostyanf14 (changes requested):

lib/engines/hcktest/hcktest.rb
akihikodaki: I'm for the former. We don't have a reason to create a snapshot for each test group while we don't create one for each test.

lib/engines/hcktest/hcktest.rb
```diff
@@ -221,20 +222,56 @@ def prepare_tests
...
+        @logger.info('Environment ready, running normal tests')
+        @tests.run(@test_list - group_tests_by_config.values.flatten)
         pause_run if @project.options.test.manual
+      end
+    end
 
-        @tests.create_project_package
+    def run_tests_with_config
+      group_tests_by_config.each do |group, tests|
+        next if tests.empty? || group == :normal
```
akihikodaki, line 253:
```suggestion
        next if tests.empty?
```

lib/setupmanagers/hckclient.rb
```diff
@@ -55,7 +55,13 @@ def restart_machine
       @tools.restart_machine(@name)
     end
 
+    def reconfigure_pool
+      @pool = @project.engine_tag
+    end
```
akihikodaki, line 60: Having `@pool = 'Default Pool'` does no longer make sense in the first place. Let's remove `@pool` and instead create a method `pool` that queries and returns the pool.

akihikodaki (changes requested):

lib/setupmanagers/hckclient.rb
```diff
@@ -46,8 +49,8 @@ def reconfigure_machine
     end
 
     def delete_machine
-      @tools.delete_machine(@name, @pool)
-      @pool = 'Default Pool'
+      @tools.delete_machine(@name, pool)
+      pool
```
akihikodaki, line 53: I don't think we need to call `pool` here.

lib/setupmanagers/hckclient.rb
```diff
@@ -18,23 +18,26 @@ def initialize(setup_manager, scope, studio, name, run_opts)
...
       @name = name
       @kit = setup_manager.kit
-      @pool = 'Default Pool'
       @logger.info("Starting client #{name}")
       @runner = setup_manager.run_client(scope, @name, run_opts)
       scope << self
     end
 
+    def pool
+      machine_in_working_pool? ? @project.engine_tag : 'Default Pool'
```
akihikodaki, line 27: You can instead use `@studio.list_pools` and return the pool containing this machine.

lib/setupmanagers/hckclient.rb
```diff
@@ -139,21 +149,27 @@ def return_when_client_up
...
+          prepare_envirenment
         end
 
-        @logger.info("Preparing client #{@name}...")
-        @project.extra_sw_manager.install_software_before_driver(@tools, @name)
-        install_drivers
-        copy_dvl
-        @project.extra_sw_manager.install_software_after_driver(@tools, @name)
         @logger.info("Configuring client #{@name}...")
         configure_machine
```
akihikodaki, line 174: Don't call `move_machine_to_pool` when `machine_in_working_pool?`.

akihikodaki (changes requested):

lib/setupmanagers/hckclient.rb
```diff
@@ -18,36 +18,42 @@ def initialize(setup_manager, scope, studio, name, run_opts)
...
       @runner = setup_manager.run_client(scope, @name, run_opts)
       scope << self
     end
 
+    def pool
+      @studio.list_pools.each do |pool|
+        return pool['name'] if pool['machines'].any? { |machine| machine['name'].eql?(@name) }
+      end
+
+      'Default Pool'
```
akihikodaki, line 31: It shouldn't return `'Default Pool'` when `list_pools` says it's not in `Default Pool`.

lib/setupmanagers/hckclient.rb
```diff
@@ -114,9 +119,16 @@ def copy_dvl
       end
     end
 
+    def machine_in_working_pool?
```
akihikodaki, line 122: This method should no longer be necessary. You can just call `pool` and check the returned value.

lib/setupmanagers/hckclient.rb
```diff
@@ -139,21 +151,28 @@ def return_when_client_up
       initialize_client_wait
     end
 
+    def prepare_envirenment
```
akihikodaki, line 154: "Environment" is not a word used elsewhere. Name it `prepare_client` or `prepare_machine`.

lib/setupmanagers/hckclient.rb
```diff
@@ -169,7 +188,7 @@ def synchronize
     end
 
     def not_ready?
-      @studio.list_pools.detect { |pool| pool['name'].eql?(@pool) }['machines']
+      @studio.list_pools.detect { |p| p['name'].eql?(pool) }['machines']
```
akihikodaki, line 191: This call of `pool` is unnecessary.

akihikodaki (changes requested):

kostyanf14:
@Jedoku HCK-CI run failed

```
00:02:40.087 F, [2024-08-28T02:50:10.086695 #148260] FATAL -- : (NoMethodError) undefined method `[]' for nil:NilClass
00:02:40.087    -- AutoHCK-PR-run/lib/setupmanagers/hckclient.rb:182:in `not_ready?'
00:02:40.087    -- AutoHCK-PR-run/lib/setupmanagers/hckclient.rb:188:in `reset_to_ready_state'
00:02:40.087    -- AutoHCK-PR-run/lib/setupmanagers/hckclient.rb:41:in `configure_machine'
00:02:40.087    -- AutoHCK-PR-run/lib/setupmanagers/hckclient.rb:168:in `block in configure'
00:02:40.087    -- AutoHCK-PR-run/vendor/bundle/ruby/3.2.0/gems/logging-2.4.0/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
```

lib/setupmanagers/hckclient.rb
```diff
@@ -116,7 +119,7 @@ def copy_dvl
 
     def machine_in_default_pool
```
akihikodaki, line 120: This method is no longer necessary; it can be simply replaced with `pool == 'Default Pool'`

lib/setupmanagers/hckclient.rb
```diff
@@ -139,21 +142,28 @@ def return_when_client_up
       initialize_client_wait
     end
 
+    def prepare_machine
+      @logger.info("Preparing client #{@name}...")
```
akihikodaki, line 146: `configure_machine` doesn't contain `@logger.info("Configuring client #{@name}...")` but `prepare_machine` does. Keep them consistent.

lib/engines/hcktest/hcktest.rb
```diff
@@ -229,20 +230,56 @@ def prepare_tests
...
         pause_run if @project.options.test.manual
+      end
+    end
 
-        @tests.create_project_package
+    def run_tests_with_config
+      group_tests_by_config.each do |group, tests|
+        next if tests.empty?
+
+        @logger.info("Preparing environment for #{group} tests")
```
akihikodaki, line 263: Log just before running tests instead of logging here because:
- `HCKClient#configure` logs before configuring anyway.
- It is inconsistent with `run_tests_without_config`, which logs just before running tests.

akihikodaki (changes requested):

akihikodaki (changes requested):
Please check the unresolved comments.

lib/engines/hcktest/hcktest.rb
akihikodaki: Please add code to raise an error for `dump_run`

lib/engines/hcktest/hcktest.rb
```diff
@@ -254,6 +292,9 @@ def run
         dump_run
 
         @project.logger.info("Find all scripts in folder: #{@project.workspace_path}")
+        @project.logger.warn(
+          'Dump mode is only for basic tests. For specific configurations (e.g., Secure Boot tests), do it manually.'
```
kostyanf14, line 296: do it manually -> update run scripts manually

kostyanf14 (approved):

lib/engines/hcktest/hcktest.rb
```diff
@@ -254,6 +292,10 @@ def run
         dump_run
 
         @project.logger.info("Find all scripts in folder: #{@project.workspace_path}")
+        @project.logger.warn(
+          'Dump mode is only for basic tests. ' \
+          'For specific configurations (e.g., Secure Boot tests), update run scripts manually'
+        )
```
akihikodaki, line 298: Split this into two `warn` calls; if the message is too long in source code, it is probably too long to fit in the terminal as well.

lib/engines/hcktest/tests.rb
```diff
@@ -6,7 +6,7 @@ module AutoHCK
   class Tests
     include Helper
 
-    HANDLE_TESTS_POLLING_INTERVAL = 10
+    HANDLE_TESTS_POLLING_INTERVAL = 30
```
akihikodaki, line 9: Is this change really necessary?

akihikodaki (changes requested):

lib/engines/hcktest/hcktest.rb
```diff
@@ -254,6 +292,10 @@ def run
         dump_run
 
         @project.logger.info("Find all scripts in folder: #{@project.workspace_path}")
+        @project.logger.warn(
+          'Dump mode is only for basic tests. ' \
+          'For specific configurations (e.g., Secure Boot tests), update run scripts manually'
```
akihikodaki, line 297: "run" is superfluous; I don't remember any code that calls the scripts "run scripts". You can just write as "scripts".

akihikodaki (approved):

PR #431, opened 2024-07-31
Add QemuHCKDevice struct

kostyanf14:


akihikodaki (changes requested):
What about hard-code everything without using `send` and `instance_variable_get` in `load_device_info`?

```Ruby
@define_variables.merge! device_info.define_variables
```

`QemuHCKDevice#define_variables` may have non-`nil` default values for convenience of the merge operation.

kostyanf14:
> What about hard-code everything without using `send` and `instance_variable_get` in `load_device_info`?

The main question is why? This will simplify the method, but not so much. We have a similar mechanism when applying platform states. In this case, hard-coding all states will be more complicated.

akihikodaki:
> > What about hard-code everything without using `send` and `instance_variable_get` in `load_device_info`?
> 
> The main question is why? This will simplify the method, but not so much. We have a similar mechanism when applying platform states. In this case, hard-coding all states will be more complicated.

Because using `send` and `instance_variable_get` will lose type safety. Platform and driver options have the same problem, but we can make them exceptions considering the additional code it requires. Types matter less to them because their merge operations are just assignments.

lib/models/qemuhck_device.rb
```diff
@@ -0,0 +1,19 @@
...
+module AutoHCK
+  module Models
+    class QemuHCKDevice < T::Struct
+      extend T::Sig
+      extend JsonHelper
+
+      const :name, String
+      const :type, T.nilable(String)
+      const :command_line, T::Array[String]
+      const :define_variables, T.nilable(T::Hash[String, String])
```
akihikodaki, line 13: Specify `default: {}` so that we don't have to make this nilable.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -161,33 +163,30 @@ def close
...
+      @define_variables = T.let({}, T::Hash[String, String])
       define_local_variables
 
       load_options(options)
       init_config
       apply_states
-      init_ports
-
-      @nm = NetworkManager.new(@id, @client_id, @machine, @logger)
-      @sm = StorageManager.new(@id, @client_id, @config, @options, @logger)
```
akihikodaki, line 172: We shouldn't move these variables away from `initialize`.

Instead, I suggest creating methods to derive `config_commands`, `pre_start_commands`, `post_stop_commands` and `define_variables` from `@device_infos` to eliminate the corresponding variables.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: Currently, `define_variables` is used in states, so we can add the `device_define_variables` method but the `define_variables` variable still must exist for `apply_state`. `config_commands`, `pre_start_commands`, `post_stop_commands` looks more device-specific.

akihikodaki: Here is my next suggestion:
* Merge `apply_states` into `init_config`. This will also eliminate `@states_config` as a bonus.
* Let `load_devices` return `T::Array[Model::QemuHCKDevice]` and assign it into `@device_infos`. This removes the mutation of `@device_infos`, an instance variable, and makes `QemuMachine` less stateful.

`QemuMachine` is too stateful so let's keep thinking about how to remove states.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -449,16 +440,16 @@ def process_storage_command
       @device_commands << devs
     end
 
+    sig { returns(T::Array[Models::QemuHCKDevice]) }
     def load_devices
-      @device_infos = []
+      @devices_list.flatten!
+      @devices_list.compact!
```
akihikodaki, line 446: I'm not for moving these into `load_devices`. The method is to read the list of devices and load the devices, not to modify the list.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: As only `load_devices` uses `@devices_list` I preferred that it will fix all duplication/etc in `@devices_list`. Also, I want to simplify `initialize` where this is possible. Maybe call `@devices_list.flatten.compact.each` instead of updating the list?

akihikodaki: It is a good idea to push state changes that happen only at initialization time to `initialize`; it will clarify the order of state changes. The class is too complicated if that complicates `initialize` too much.

In this case, I believe `@devices_list` should be typed as `T::Array[String]`, and arrays or `nil` shouldn't be pushed into `@devices_list`, which ensures the array is always in a form suited for consuming. This also applies to other variables.

However, I don' request to make such a change now. Keeping the status quo by leaving `@devices_list.flatten!` and `@devices_list.compact!` in `initialize` is enough for this pull request.

akihikodaki (approved):

PR #440, opened 2024-08-08
Redirect stdout to stderr in bin/ns_unshared

akihikodaki:


lib/project.rb
```diff
@@ -71,8 +71,8 @@ def log_exception(exception, level)
     def init_multilog(verbose)
       @temp_pre_logger = StringIO.new
       @pre_logger = MonoLogger.new(@temp_pre_logger)
-      @stdout_logger = MonoLogger.new($stdout)
-      @logger = MultiLogger.new(@pre_logger, @stdout_logger)
+      @stderr_logger = MonoLogger.new($stderr)
+      @logger = MultiLogger.new(@pre_logger, @stderr_logger)
```
kostyanf14, line 75: why do we need to use stderr instead of stdout?

kostyanf14: In this case please update lib/config/sentry.rb:
```
 MonoLogger.new($stdout).warn('SENTRY_TRANSPORT_SSL_CA environment specified, but file does not exist')
 ```

Comment (akihikodaki) [reply lib/project.rb:75]: Done tree-wide search with commit a241bce6ce3b99926c4d5f96bcc5d20cef74e272; actually the error with `SENTRY_TRANSPORT_SSL_CA` was preventing running AutoHCK for Vitalii.

File: bin/triggers_check [hunk 2]
```diff
@@ -85,7 +85,7 @@ module AutoHCK
     cli = CLI.new
     cli.parse(ARGV)
 
-    @logger = MonoLogger.new($stdout)
+    @logger = MonoLogger.new($stderr)
     @logger.level = cli.debug ? 'DEBUG' : 'INFO'
```
kostyanf14, line 89: The same logger is used at the end to write: `Some drivers are changed, running tests` or `Any drivers aren't changed, not running tests`. Do you think this log is an actual product of the operation or an actual product is it the exit code?

akihikodaki: It is the exit code in this case. We don't expect anybody to parse that log.

kostyanf14 (approved):

PR #451, opened 2024-08-23
SVVP: Allow Verify IOMMU presence and configuration test 

kostyanf14:


lib/setupmanagers/qemuhck/devices/iommu.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "iommu",
   "command_line": [
-    "-device intel-iommu,intremap=on,aw-bits=48,caching-mode=on,device-iotlb=on"
+    "-device intel-iommu,intremap=on,aw-bits=48,caching-mode=off,eim=on"
```
akihikodaki, line 4: For `device-iotlb`, QEMU documentation says:
> This enables device-iotlb capability for the emulated VT-d device.  So far virtio/vhost should be the only real user for this parameter, paired with ats=on configured for the device.

`iommu_device_param` still has `ats=on`, which probably does not make sense.

svvp.json
```diff
@@ -26,7 +26,6 @@
     "Profile Interrupt Test",
     "__Windows_Server_2022_reject_test_names",
     "UEFI GOP mode test",
-    "Verify IOMMU presence and configuration",
```
akihikodaki, line 29: I don't think it is a good idea to remove this while IOMMU is not always enabled.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/devices/iommu.json
kostyanf14: Added device-iotlb.

svvp.json
kostyanf14: As for me `reject_test_names` is a list of unsupported tests at all. Yes, some platform will fail IOMMU presence tests but the simmilar situation with a lot of other tests. `Check SMBIOS *` can fail with BIOS due to bug in QEMU/Windows. `Crashdump Support Test` will fail when client does not have internet access. `Hardware Security Testability Interface Test` requied new OVMF with HSTI table. But in general with proper CLI, platform, OVMF, etc all of these tests will pass. 

SVVP testing is very specific and user should use the proper platform if want to pass all tests.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -318,7 +318,7 @@ def memory_replacement_map
 
     def options_replacement_map
       {
-        '@machine_options@' => @machine_options.join(','),
+        '@machine_options@' => @machine_options.concat(device_machine_options).join(','),
```
akihikodaki, line 321: Surprisingly `Array#concat` is a destructive operation, which means `@machine_options` will be modified. `+` operator should be used instead to avoid such a side effect.

akihikodaki (changes requested):

svvp.json
akihikodaki: The requirements of `Check SMBIOS *`, `Crashdump Support Test`, `Hardware Security Testability Interface Test` are not under the control of AutoHCK so it makes sense to require the user to fulfill them.

The situation is different for IOMMU. It is AutoHCK that controls the device to enable so AutoHCK should always enable IOMMU for SVVP if it is going to test `Verify IOMMU presence and configuration`.

akihikodaki: That is due to a bug in QEMU/Windows. AutoHCK will run SVVP tests properly if the bug gets fixed. However `Verify IOMMU presence and configuration` will not run unless AutoHCK enables IOMMU. In this sense, it is AutoHCK's responsibility to get it to work, and we should enable it when we satisfy the responsibility.

Isn't it a matter of adding `iommu` to `drivers` by the way?

kostyanf14: maybe better to add `viommu_state: true` to svvp config?

akihikodaki: That sounds good to me.

kostyanf14: in any case, we have a problem with re-adding devices. If `iommu` will be specified in `devices_list` and `viommu_state: true` is in the platform, AutoHCK adds iommu again.

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: done

svvp.json
kostyanf14: done

lib/engines/hcktest/hcktest.rb
```diff
@@ -118,7 +118,16 @@ def target
...
     def self.platform(logger, options)
       platform_name = options.test.platform
-      Json.read_json("#{PLATFORMS_JSON_DIR}/#{platform_name}.json", logger)
+      platform = Json.read_json("#{PLATFORMS_JSON_DIR}/#{platform_name}.json", logger)
+
+      if options.test.svvp
+        svvp_info = Models::SVVPConfig.from_json_file(SVVP_JSON, logger)
+
+        platform['viommu_state'] ||= svvp_info.viommu_state
+        logger.debug("Platform updated from SVVP config: #{platform}")
```
akihikodaki, line 127: I don't think this log is useful unless you are modifying this particular method.

akihikodaki (changes requested):

akihikodaki (approved):

PR #457, opened 2024-08-28
POC: QemuHCK: Load coresponding IOMMU device

kostyanf14:


kostyanf14:
@Jedoku 

As you have AMD CPU can you please test the PR

lib/setupmanagers/qemuhck/devices/amd-iommu.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "iommu",
+  "command_line": [
+    "-device amd-iommu,intremap=on,device-iotlb=on,xtsup=on"
```
Jedoku, line 4: `CmdRun PID:59231: stderr: qemu-system-x86_64: -device amd-iommu,intremap=on,device-iotlb=on,xtsup=on: Property 'amd-iommu.xtsup' not found`

Jedoku: It works well without `xtsup=on`

kostyanf14: This is a new flag that was added by the AMD team. Thanks for testing.

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,29 @@
...
+  class IommuDevice
+    extend T::Sig
+
+    def initialize(logger)
+      @logger = logger
+    end
+
+    sig { returns(Models::QemuHCKDevice) }
+    def qemu_hck_device
+      vendor = Sys::CPU.processors[0].vendor_id
```
akihikodaki, line 16: The gem is too complicated just to get the vendor ID. You can do: `File.read('/proc/cpuinfo').match(/^vendor_id\s*:\s*(.*)/)[1]`

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'sys/cpu'
+
+module AutoHCK
+  class IommuDevice
+    extend T::Sig
+
+    def initialize(logger)
+      @logger = logger
+    end
```
akihikodaki, line 12: I think it is simpler to have a `module` named `QemuHCK::Devices` and have a class method named `iommu`. This removes the need to assign an instance variable.

akihikodaki (changes requested):

akihikodaki:
> @akihikodaki
> 
> What do you think about this implementation of dynamic IOMMU for AMD/Intel? Do you have other suggestions?

I think the idea of having a "dynamic" device is good. I wrote a few comments on implementation details.

lib/setupmanagers/qemuhck/devices/iommu.rb
kostyanf14: Do you expect to call `device.public_send("iommu")` or have a list of condition `if dev == 'iommu' then device.iommu` ?

akihikodaki: You can still keep the logic you already have in `read_dynamic_device`. Just replace `AutoHCK.const_get` with `QemuHCK::Devices.public_send`.

kostyanf14:
Moved to https://github.com/HCK-CI/AutoHCK/pull/469

PR #469, opened 2024-09-15
Add support for AMD IOMMU

Jedoku:


lib/auxiliary/multi_logger.rb
```diff
@@ -3,10 +3,10 @@
 # AutoHCK module
 module AutoHCK
   # MultiLogger class
-  class MultiLogger
+  class MultiLogger < Logger
```
kostyanf14, line 6: Why do we need this inheritance?

lib/auxiliary/multi_logger.rb
```diff
@@ -3,10 +3,10 @@
...
     attr_reader :level
 
-    def initialize(*loggers)
+    def initialize(*loggers) # rubocop:disable Lint/MissingSuper
```
YanVugenfirer, line 9: Please move this change to separate commit

As a rule of thumb please always have cosmetics, unrelated to main topic changes, and etc as a separate commit

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module AutoHCK
+  class QemuHCK
+    module Devices
+      extend T::Sig
+
+      sig { params(logger: AutoHCK::MultiLogger).returns(Models::QemuHCKDevice) }
+      def self.iommu(logger)
+        cpuinfo = File.read('/proc/cpuinfo')
+        vendor = T.must(cpuinfo.match(/^vendor_id\s*:\s*(\S+)/))[1]
```
akihikodaki, line 12: Check if `cpuinfo.match(/^vendor_id\s*:\s*(\S+)/)` if `nil` just as you do for `[1]` below instead of calling `T.must`.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,25 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      file_full_path = File.realpath("#{DEVICES_JSON_DIR}/#{device}.rb")
+      require file_full_path
+
+      raise InvalidConfigFile, "#{device} does not exist" unless QemuHCK::Devices.respond_to?(device)
+
+      device_obj = QemuHCK::Devices.public_send(device, @logger)
```
akihikodaki, line 383: Let's call it `info` as done elsewhere. `obj` is meaningless as everything in Ruby is an object and shouldn't be used for names.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,25 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      file_full_path = File.realpath("#{DEVICES_JSON_DIR}/#{device}.rb")
+      require file_full_path
```
akihikodaki, line 379: Let's call `File.exist?` here and return `nil` if false instead of calling `File.exist?` in `read_device` so that we can use `file_full_path`.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,25 @@ def full_replacement_map
...
+
+      raise InvalidConfigFile, "#{device} does not exist" unless QemuHCK::Devices.respond_to?(device)
+
+      device_obj = QemuHCK::Devices.public_send(device, @logger)
+
+      raise InvalidConfigFile, "#{device} does not exist" unless device_obj
```
akihikodaki, line 385: This check is unnecessary. Add `nil` checks only where you know the value can be `nil`; otherwise we will need add `nil` checks everywhere.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,25 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      file_path = Pathname.new(DEVICES_JSON_DIR).join("#{device}.rb")
```
akihikodaki, line 378: Call `realpath` here instead of after `File.exist?` so that we don't have to resolve symlinks twice.

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,27 @@
...
+
+        case vendor
+        when 'GenuineIntel'
+          Models::QemuHCKDevice.from_json_file("#{__dir__}/intel-iommu.json", logger)
+        when 'AuthenticAMD'
+          Models::QemuHCKDevice.from_json_file("#{__dir__}/amd-iommu.json", logger)
+        else
+          logger.fatal("Unknown CPU vendor: #{vendor}")
+          exit 1
+        end
```
akihikodaki, line 23: This will cause logging `Unknown CPU vendor` even when `vendor_match` is `nil`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
Jedoku: we cant do that, _.realpath_ assume that file exists

akihikodaki: Then let's rescue `Errno::ENOENT` for `realpath` instead of having another `File.exist?` check.

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,34 @@
...
+        cpuinfo = File.read('/proc/cpuinfo')
+        vendor_match = cpuinfo.match(/^vendor_id\s*:\s*(\S+)/)
+
+        vendor =
+          if vendor_match
+            vendor_match[1]
+          else
+            logger.fatal('Could not determine CPU vendor')
+            exit 1
+          end
```
akihikodaki, line 20: Simpler:

```suggestion
        if vendor_match.nil?
          logger.fatal('Could not determine CPU vendor')
          exit 1
        end

        vendor = vnedor_match[1]
```

akihikodaki (changes requested):
Please don't call `exit` in an ad-hoc manner; raise an error and let the caller handle it instead.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,23 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      require Pathname.new(DEVICES_JSON_DIR).join("#{device}.rb").realpath.to_s
+
+      raise InvalidConfigFile, "#{device} does not exist" unless QemuHCK::Devices.respond_to?(device)
+
+      QemuHCK::Devices.public_send(device, @logger)
+    rescue Errno::ENOENT
```
akihikodaki, line 383: Don't rescue the whole method. It hides an error caused by `Models::QemuHCKDevice.from_json_file("#{__dir__}/amd-iommu.json", logger)`, for example.

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,33 @@
...
+      extend T::Sig
+
+      sig { params(logger: AutoHCK::MultiLogger).returns(Models::QemuHCKDevice) }
+      def self.iommu(logger)
+        cpuinfo = File.read('/proc/cpuinfo')
+        vendor_match = cpuinfo.match(/^vendor_id\s*:\s*(\S+)/)
+
+        if vendor_match.nil?
+          logger.fatal('Could not determine CPU vendor')
+          exit 1
```
akihikodaki, line 16: Please `raise` instead of `exit`. It's the caller to decide how to handle an error.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,25 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      begin
+        require Pathname.new(DEVICES_JSON_DIR).join("#{device}.rb").realpath.to_s
```
akihikodaki, line 379: We shouldn't rescue `require`; `require` may raise `Errno::ENOENT` because of the code written in the required file.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/devices/iommu.rb
```diff
@@ -0,0 +1,33 @@
...
+    module Devices
+      extend T::Sig
+
+      sig { params(logger: AutoHCK::MultiLogger).returns(Models::QemuHCKDevice) }
+      def self.iommu(logger)
+        cpuinfo = File.read('/proc/cpuinfo')
+        vendor_match = cpuinfo.match(/^vendor_id\s*:\s*(\S+)/)
+
+        if vendor_match.nil?
+          logger.fatal('Could not determine CPU vendor')
```
akihikodaki, line 15: These logs are unnecessary. The caller should properly log the exception.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -374,9 +374,27 @@ def full_replacement_map
                          @define_variables)
     end
 
+    def read_dynamic_device(device)
+      begin
+        device_file = Pathname.new(DEVICES_JSON_DIR).join("#{device}.rb").realpath.to_s
```
akihikodaki, line 379: - It is better to name it like `device_path` because it is not a `File` but a path to one.
- Let's move `to_s` out of `rescue`. It's better to have less code in `begin` - `rescue`.

akihikodaki (changes requested):

akihikodaki (approved):

kostyanf14 (approved):

PR #476, opened 2024-09-24
Add support for S3-compatible storage in Result Uploader

zjmletang:
To upload files using Alibaba Cloud OSS, you must first have an Alibaba Cloud OSS account, which includes your Access Key (AK) and Secret Key (SK). Once you have these, please follow the steps below for configuration:

1. Create an OSS Bucket: Start by setting up a new OSS bucket in your Alibaba Cloud account.
2. Configure the alicloudoss.json File: Update this file with your bucket name and endpoint.
3. Set Environment Variables: Configure your AK and SK as environment variables by executing:
   export AUTOHCK_ALICLOUDOSS_ACCESS_KEY_ID=\<AK\>
   export AUTOHCK_ALICLOUDOSS_CLIENT_SECRET=\<SK\>
After completing these steps, you'll be ready to upload files to Alibaba Cloud OSS.

for the problem description
please visit HCK-CI#471
    
Signed-off-by: Zhang JianMing zhangjianming.zjm@alibaba-inc.com


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'rtoolsHCK', git: 'https://github.com/HCK-CI/rtoolsHCK.git', tag: 'v0.5.1'
 gem 'rubyzip'
 gem 'sentry-ruby'
 gem 'sorbet-runtime'
+gem 'aliyun-sdk'
```
kostyanf14, line 20: As you add a new gem and we use Sorbet for type checking, please regenerate Sorbet RBIs and commit with Gemfile update. `bundle exec tapioca gem` - to regenerate

lib/resultuploaders/alicloudoss/alicloudoss.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    include Helper
+
+    attr_reader :url
+
+    def initialize(project)
+      @tag = project.engine_tag
+      @timestamp = project.timestamp
+      @logger = project.logger
+      puts "@logger is  #{@logger}"
```
kostyanf14, line 20: Looks like leave after some debugging, please remove it.

kostyanf14 (changes requested):
Please review and fix Rubocop code scanning error too.

lib/resultuploaders/alicloudoss/alicloudoss.rb
zjmletang: ok

akihikodaki (changes requested):
First, thank you for your contribution.

Please consider using AWS SDK instead; it should give chances to support many S3-compatible storages, including Alibaba Cloud OSS. If you definitely need a feature specific to Alibaba Cloud OSS, please explain it.

zjmletang:
> First, thank you for your contribution.
> 
> Please consider using AWS SDK instead; it should give chances to support many S3-compatible storages, including Alibaba Cloud OSS. If you definitely need a feature specific to Alibaba Cloud OSS, please explain it.

@akihikodaki , Thank you for your valuable suggestion! I chose the Alibaba Cloud SDK primarily because I work with Alibaba Cloud, which makes maintenance and troubleshooting more convenient for me. I hadn't fully considered the potential benefits of using the AWS SDK, as I usually select the SDK that corresponds to the cloud service I'm utilizing.

I do have a question regarding the definition of "result uploader." Currently, we are using Dropbox, and my initial understanding was that each storage service would have its own dedicated "result uploader," such as Dropbox, Alibaba Cloud OSS, AWS Storage, etc. Is this understanding correct? In other words, even though they are all S3-compatible storages, will there still be individual result uploaders for different products from various companies? I would appreciate your insight on whether you envision S3-compatible storages as being treated as a single result uploader.

kostyanf14:
@zjmletang 

>I would appreciate your insight on whether you envision S3-compatible storages as being treated as a single result uploader.

We agree with the idea of having one S3-compatible result uploader with the ability to configure it for AWS, Alibaba Cloud, or even open-source storage (e.g., minio).

zjmletang:
> We agree with the idea of having one S3-compatible result uploader with the ability to configure it for AWS, Alibaba Cloud, or even open-source storage (e.g., minio).

@kostyanf14  Got it. I have turned the patch into a draft as I need some time to modify and test it

zjmletang:
@kostyanf14  Could you please reopen my patch? I accidentally closed it while resetting the branch, and I didn’t see the option to reopen it afterward. I’m not sure if this is due to permission issues.

YanVugenfirer:
@zjmletang I reopened

kostyanf14:
@zjmletang Can I make PR ready for review?

zjmletang:
@kostyanf14. Not ready yet; I haven't tested it thoroughly myself.  I'll notify you once I finish the testing

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,111 @@
...
+
+# AutoHCK module
+module AutoHCK
+  # S3compatible class
+  class S3compatible
+    CONFIG_JSON = 'lib/resultuploaders/s3compatible/s3compatible.json'
+
+    include Helper
+
+    attr_reader :url
```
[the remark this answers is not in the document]
zjmletang: Yes, this part is not fully completed yet. Since there is no concept of folders in S3, directly accessing a folder URL may not be feasible. I see two methods that could potentially replace accessing the folder URL:

1. Hosting the bucket as a static website, which requires a lot of upfront configuration. see https://docs.aws.amazon.com/AmazonS3/latest/userguide/HostingWebsiteOnS3Setup.html
2. Creating an index.html file in the folder, which the URL will point to. This index.html will store the URLs of the actual files, and it will be updated every time files are uploaded or deleted.
image


Currently, I plan to choose option 2 (the code will be uploaded later). What do you think?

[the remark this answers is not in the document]
kostyanf14: Looks good according to S3 behavior. @akihikodaki What do you think?

zjmletang:
@kostyanf14 , I am ready on my side. Please see the configuration document below

To upload files using S3-Compatible storage (supports AWS S3, Alibaba Cloud OSS, etc.), please follow the steps below for configuration:
1. Create an Account
  For AWS S3: Sign up for an AWS account and obtain your Access Key (AK) and Secret Key (SK).
  For Alibaba Cloud OSS: Similarly, create an Alibaba Cloud account and get your Access Key (AK) and Secret Key (SK).
2. Create a Storage Bucket
   Log in to your cloud provider's console (AWS or Alibaba Cloud) and create a new storage bucket.
3. Set Bucket Permissions
   Ensure that your bucket is configured to have public read permissions.
4. Configure s3compatible.json File
  Update the s3compatible.json file in your project directory with the following content:
{
    "bucket_name": "<your_bucket_name>",
    "region": "<your_region>",
    "endpoint": "<your_endpoint>"
}
   Example Configurations:
     For AWS S3:
{
       "bucket_name": "auto-hck-result",
        "region": "us-east-1",
        "endpoint": "s3.amazonaws.com"
}
    For Alibaba Cloud OSS:
{
    "bucket_name": "auto-hck-result",
    "region": "oss-cn-hangzhou",
    "endpoint": "oss-cn-hangzhou.aliyuncs.com"
}
   For specific values of region and endpoint, please refer to the official documentation of your cloud provider.
5. Set Environment Variables
   Configure your AK and SK as environment variables by executing the following commands:
export AUTOHCK_S3COMPATIBLE_ACCESS_KEY_ID=\<AK\>
export AUTOHCK_S3COMPATIBLE_SECRET_ACCESS_KEY=\<SK\>

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+    private
+
+    def update_index_file
+      title = @path.split('/').last
+      html_content = build_html_content(title)
+      html_content += list_objects
+      save_index_file(html_content)
+    end
+
+    def build_html_content(title)
```
kostyanf14, line 136: I think better to use ERB instead. See example `lib/templates/report.html.erb` and `lib/engines/hcktest/tests.rb` function `generate_report`

zjmletang: Indeed, this is more appropriate. I'll make some changes. Thank you for the guidance!

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
akihikodaki: I suspect you will want to see `report.html` instead of the file list. `report.html` may include the file list or link to one.

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-s3'
+require './lib/auxiliary/json_helper'
+
+# AutoHCK module
```
akihikodaki, line 6: Please remove this comment as it is redundant. Context: https://github.com/HCK-CI/AutoHCK/commit/7fe2b623c1b7ba9dbb310ba2f3ae7190039a9fab

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-s3'
+require './lib/auxiliary/json_helper'
```
akihikodaki, line 4: Please remove this as they are now autoloaded. See: https://github.com/HCK-CI/AutoHCK/pull/407

akihikodaki (changes requested):
I prefer it to be named S3 instead of S3compatible for conciseness. Anything compatible with S3 should work with an uploader for S3 anyway so "compatible" adds no semantic value.

lib/resultuploaders/s3compatible/s3compatible.rb
zjmletang: ok

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    def connect
+      handle_exceptions(__method__) do
+        @s3_resource = Aws::S3::Resource.new(
+          access_key_id: @access_key_id,
+          secret_access_key: @secret_access_key,
+          region: @region, endpoint: "https://#{@endpoint}",
+          force_path_style: true # Necessary for compatibility with S3-compatible services
+          # such as Alibaba Cloud OSS, which do not support virtual-hosted style.
+          # Forces path-style access to ensure proper functionality.
```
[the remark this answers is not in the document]
zjmletang: I considered making this configurable before, but I was concerned that this parameter might have a learning curve for those who are not familiar with S3. Based on my tests, I found that using path-style can provide better compatibility with more S3 storage options (at least AWS and Alibaba Cloud both support it, and Alibaba Cloud OSS only supports path-style), which is why I decided to hard-code it in the implementation. Of course, making it configurable would definitely be more flexible. What do you think?

lib/resultuploaders/s3compatible/s3compatible.rb
zjmletang: ok.

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+    # Future enhancements may include more robust error handling strategies.
+    def handle_exceptions(where)
+      yield
+    rescue Aws::Errors::ServiceError => e
+      @logger.error("S3Compatible Error in #{where}: #{e.message}")
+      false
+    rescue StandardError => e
+      @logger.error("S3Compatible General error in #{where}: #{e.message}")
+      false
+    end
```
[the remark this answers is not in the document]
zjmletang: I see your point. Indeed, there is currently no recovery mechanism implemented. I've noticed that the S3 API itself has some built-in retry and recovery features, so the need for additional recovery mechanisms seems limited. The primary purpose of this implementation is to alert users to preliminary errors, such as misconfigured buckets or incorrect access keys, which can still provide valuable guidance and aid during troubleshooting.
I have combined the rescue for Aws::Errors::ServiceError and StandardError. Is it correct to implement it this way?
def handle_exceptions(where)
  yield
rescue StandardError => e
  @logger.error("S3Compatible Error in #{where}: #{e.message}")
  false
end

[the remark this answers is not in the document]
akihikodaki: Doesn't AutoHCK automatically handle the error? If not, perhaps it is better to do it in code common with the other uploader (Dropbox).

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
zjmletang: @akihikodaki ，Thank you for your feedback! I'm a bit unclear about your point. Could you please elaborate a bit more? When you mention report.html, are you referring to result.html（see below)? My intention is indeed to provide a file list, which refers to the URL of the folder containing the test results. Since S3 doesn’t have a concept of folders, I used 'file list' to represent this idea. By accessing that URL, we can retrieve all the results from this test.
image

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
zjmletang: I understand now. I'll make some changes. Thank you

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
akihikodaki: Yes, I referred to that file. I see the utility of the file list, but I suspect perhaps you may want to see `results.html` first, and navigate to the file list from there.

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+    def save_index_file(html_content)
+      obj = @bucket.object("#{@path}/#{INDEX_FILE_NAME}")
+      obj.put(body: html_content, content_type: 'text/html')
+    end
+
+    # @note It is required that the bucket has public read permissions set.
+    # This ensures that the generated URL will be accessible without authentication.
+    def generate_url(object)
+      "https://#{@bucket_name}.#{@endpoint}/#{object}"
+    end
```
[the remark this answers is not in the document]
zjmletang: The main reason I don’t use public_url is that it is compatible with AWS but not with Alibaba Cloud OSS. I noticed that the URL format generated by public_url is like this: https://#{@endpoint}/@bucket_name}#{object}, whereas the URL generation format for Alibaba Cloud is https://#{@bucket_name}.#{@endpoint}/#{object}. These two formats are different. The latter format that I piece together also works for AWS. What do you think?"

alibaba cloud oss detais see 
https://www.alibabacloud.com/help/en/oss/user-guide/how-to-obtain-the-url-of-a-single-object-or-the-urls-of-multiple-objects#81455fb05dw1a
image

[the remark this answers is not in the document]
akihikodaki: That's strange. You specify `force_path_style`, but you say this needs to return a hosted-style URL.
The only logical explanation is that you need to use the path style for uploading and the hosted style for downloading, which is awkward.

I checked the documentation and it seems Alibaba Cloud OSS *always* requires to use the hosted-style.

Overview - Object Storage Service - Alibaba Cloud Documentation Center
https://www.alibabacloud.com/help/en/oss/developer-reference/overview-24
> For security reasons, you can use only virtual hosted-style URLs instead of path-style URLs when you initiate RESTful API requests. A virtual hosted-style URL contains the bucket name.

There is also documentation that explains how to use AWS's SDK with Alibaba Cloud OSS. Unfortunately, there is no example of Ruby, but at least the example for Java has: `withPathStyleAccessEnabled(false)`.

A possible explanation is that `force_path_style` is never intended to be used with Alibaba Cloud OSS, but it is functioning as a workaround for another bug during upload. In that case, it's better to investigate that bug.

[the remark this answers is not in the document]
zjmletang: I'm sorry, I double-checked, and you are indeed correct; Alibaba Cloud OSS does not actually require force_path_style to be set to true. I might have confused it during debugging. Thank you for the reminder

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
zjmletang: Okay, thank you, I understand your point now. My original intention was indeed to provide the URL of the folder directly. If users need to access result.html, they can obtain it through the URL of this folder. If I change the URL to point to result.html, I will encounter an issue that I haven't figured out how to address appropriately. Specifically, result.html is only generated during the “bin/auto_hck test", but not during "bin/auto_hck install". Additionally, if an exception occurs during the HCK test, result.html might not be generated, and users may want to refer to other log files to identify the issue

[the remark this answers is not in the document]
akihikodaki: Fair point about install. Let's keep the index.

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+                    margin: 5px 0;
+                }
+                a {
+                    text-decoration: none;
+                    color: #007bff;
+                }
+                a:hover {
+                    text-decoration: underline;
+                }
+            </style>
```
akihikodaki, line 165: Please use Bootstrap as `lib/templates/report.html.erb` does for consistency.

lib/resultuploaders/s3compatible/s3compatible.rb
```diff
@@ -0,0 +1,199 @@
...
+        </head>
+        <body>
+            <h1>#{title}</h1>
+            <ul>
+      HTML
+    end
+
+    def list_objects
+      objects_html = ''
+      objects = @bucket.objects(prefix: @path)
```
akihikodaki, line 175: Please do not use `@bucket.objects(prefix: @path)`, which incurs extra traffic. Remember objects when uploading instead.

akihikodaki (changes requested):

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
zjmletang: @akihikodaki I looked into it, and if exceptions are not caught here and are instead handled directly by AutoHCK, the entire program will exit if an error occurs. I noticed that Dropbox handles different errors in different ways; some exceptions are thrown, while others are just logged. To be honest, I don't fully understand the logic behind this handling and whether it relates to certain characteristics of Dropbox itself.

Currently, my approach is to catch all exceptions thrown by S3 but only log them. Do you think this needs some improvement? How could it be improved to be more appropriate?

@akihikodaki @kostyanf14  Aside from this review, I have already made the necessary changes to the others. Please take a moment to look at them as well.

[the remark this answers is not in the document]
akihikodaki: `handle_exceptions` can be removed. The default behavior, which is to exit the program with a non-zero status code, is the correct way to handle the error in my opinion. That allows the user to notice the uploaded results are corrupted and to fix it.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+      @secret_access_key = ENV.fetch('AUTOHCK_S3_SECRET_ACCESS_KEY')
+
+      @region = @config['region']
+      @bucket_name = @config['bucket_name']
+      @endpoint = @config['endpoint']
+
+      @s3_resource = nil
+      @bucket = nil
+
+      @uploader_urls = []
```
akihikodaki, line 43: - Use `SortedSet`, which sorts elements and allows finding an element with less cost.
- Remove the `uploader_` prefix. It is obvious that it belongs to an uploader.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+      handle_exceptions(__method__) do
+        @s3_resource = Aws::S3::Resource.new(
+          access_key_id: @access_key_id,
+          secret_access_key: @secret_access_key,
+          region: @region, endpoint: "https://#{@endpoint}"
+        )
+        @bucket = @s3_resource.bucket(@bucket_name)
+        @logger.info("S3 bucket connected: #{@bucket_name}")
+        true
+      end
```
akihikodaki, line 75: Perform this operation in (a method called from) `initialize`. This only creates objects and performs no network activity like creating a TCP connection, which the method name implies.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    # This method is intentionally left blank to maintain the interface,
+    # as S3-compatible does not require this function. It prevents external calls
+    # from causing errors while adhering to the expected API structure.
+    def ask_token; end
+
+    def connect
+      handle_exceptions(__method__) do
+        @s3_resource = Aws::S3::Resource.new(
```
akihikodaki, line 67: Convert this instance variable to a local one to reduce the state complexity.

lib/templates/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <title><%= title %></title>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" rel="stylesheet">
+</head>
+<body>
+    <div class="container">
+        <h1 class="text-center mt-5"><%= title %></h1>
+        <ul class="list-group mt-3">
+            <% urls.each do |url| %>
+                <% file_name = url.split('/').last %>
+                <% friendly_file_name = URI.decode_www_form_component(file_name) %>
```
akihikodaki, line 15: Use `Aws::S3::Object#key`.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+    def close; end
+
+    private
+
+    def index_data
+      {
+        'title' => @path.split('/').last,
+        'urls' => @uploader_urls
+      }
+    end
```
akihikodaki, line 133: Merge this into `generate_index`; this method is too trivial.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+    # Provides read access to the URL for the index.html file.
+    #
+    # In the context of S3 compatibility, there is no concept of folders.
+    # The `url` will point to an index.html file that records the paths of all files.
+    # This allows `index.html` to be displayed directly in HTML format.
+    #
+    # Note: While AWS S3 supports this behavior of rendering HTML files directly,
+    # not all S3-compatible storage services do. For example, Alibaba Cloud OSS
+    # has security restrictions on its default domain that may require files to be
+    # downloaded rather than rendered directly, thus preventing HTML files from being displayed.
```
akihikodaki, line 22: The documentation says it is possible to host a static website:
https://www.alibabacloud.com/help/en/oss/configure-static-website-hosting

akihikodaki (changes requested):

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
kostyanf14: @akihikodaki I don't agree. Results uploader is not a main feature. Dropbox sometimes fails and we don't have a URL or some files but CI in general passes. If we missing one file this is not critical but terminate CI just because of a network issue is a bad idea.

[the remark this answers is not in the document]
akihikodaki: Shouldn't the CI be retried when the uploader fails? It may not be that troublesome to retry the CI if uploader failures are rare.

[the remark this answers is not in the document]
kostyanf14: Can try if this is possible. In Dropbox we try several times to reconnect or perform operation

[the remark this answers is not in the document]
akihikodaki: For now, I suggest removing `handle_exceptions` for S3 but keeping `handle_exceptions` for Dropbox.
S3 and similar services have great availability and can process so many requests. An exception is likely to be an indication of programming error or misconfiguration.
I'm not that confident about Dropbox though so let's keep `handle_exceptions` for Dropbox; we can revisit it whenever necessary.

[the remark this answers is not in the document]
YanVugenfirer: @akihikodaki Sorry, I disagree.

AutoHCK is no longer just a developer tool. It runs in production deployment. Killing a CI run because of the uploaded exception is unacceptable.
With DropBox, it was a case before we added an exception handler in any case, and it was extremely annoying to find from time to time that the test failed just because of some throttling or unavailability of the service. 
Log the error if it happens. We also have integration with Sentury - trigger event in Sentury in case of S3 exception.

lib/resultuploaders/s3/s3.rb
zjmletang: Building on your previous comment (suggesting the use of Aws::S3::Object#key instead of the URL I used), I have a clearer understanding of your point. You are suggesting that we access the bucket by configuring it as a static website, which aligns with the solution I previously referred to as option 1. However, my current implementation is actually option 2, where I simulate a static webpage using index.html rather than using a true static webpage. The advantage of this approach is that users do not need to perform any additional configuration, but the trade-off is that they can only access it through the S3 default domain.

image

Regarding the comment about the access restrictions with Alibaba Cloud, I am specifically addressing the limitations related to their default domain. For files with an .html extension, Alibaba Cloud, for security reasons, does not open them in text/html format in the browser; instead, it triggers a download. You can find more details in the following links: [Alibaba Cloud Documentation 1](https://www.alibabacloud.com/help/en/oss/user-guide/overview-71?spm=a2c63.p38356.0.0.6d6836b3n6E2YQ) and [Alibaba Cloud Documentation 2](https://www.alibabacloud.com/help/en/oss/user-guide/how-to-ensure-an-object-is-previewed-when-you-access-the-object?spm=a2c63.p38356.0.0.2a5512b5Bu7J3v). However, if we aim for full support for Alibaba Cloud, it may indeed be more appropriate to choose option 1, allowing users to configure a custom domain and thereby bypass this restriction. Of course, if we switch to option 1, some modifications to my current code will be necessary. If my understanding is incorrect, please let me know.

zjmletang: However, I must say that using option 1 can indeed be somewhat cumbersome, as it requires users to register a custom domain, which involves additional costs and processes like record-filing. If users do not opt for a custom domain, they will still face the same issues associated with the default domain used in option 2. If one can accept that accessing Alibaba Cloud OSS URLs via the browser results in downloads instead of displaying them in HTML format, option 2 also seems like a viable choice. What do you think? @akihikodaki @kostyanf14

lib/resultuploaders/s3/s3.rb
zjmletang: ok

lib/resultuploaders/s3/s3.rb
zjmletang: Okay, indeed

lib/resultuploaders/s3/s3.rb
zjmletang: Thank you for your suggestion! My original intention was to display the items in the order they were uploaded, similar to how Dropbox does it (I noticed they don’t apply any additional sorting). I’m a bit concerned that using SortedSet might change the display order from what I expected.I
I’d love to hear your thoughts on this and discuss how we can best achieve the desired display order

lib/resultuploaders/s3/s3.rb
zjmletang: Thank you for your feedback! I created the index_data method to mimic the report_data function in tests.rb, with the thought that having a dedicated method might provide clarity when passing data to ERB. I also wanted to allow for future enhancements more easily. However, I appreciate your perspective on simplifying the code, and I'm open to discussing the best approach for both clarity and maintainability moving forward!

lib/resultuploaders/s3/s3.rb
kostyanf14: @zjmletang @akihikodaki 
I agree with option 2 to have index.html. Hosting a static website is more complicated and requires specific configuration. I want to leave this to a user who will use the S3 uploader. If in the future we get other requests from the community/customers we can update this uploader.

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
akihikodaki: I still think an upload failure requires the user's attention. S3 and similar services have high availability and can simultaneously serve multiple users while a Dropbox bucket is presumably designed for a few users. Both of throttling and unavailability should be rare conditions.

lib/resultuploaders/s3/s3.rb
akihikodaki: I suggested using `Aws::S3::Object#key` for the user-visible label instead of parsing URL; it is irrelevant with this comment.
I thought the static website hosting allows using the default domain as it has examples such as `examplebucket.oss-cn-hangzhou.aliyuncs.com`, but apparently that's not true. Let's leave the comment.

lib/resultuploaders/s3/s3.rb
akihikodaki: The consistency with Dropbox makes sense. Let's keep it an Array.

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
YanVugenfirer: Even if they are rare, when it happens in production, especially when CI is part of the pipeline, it is very destructive not to handle errors that are coming from auxiliary features. 

Better not crash CI and mark it pass or fail based on the test results, and let the user to handle the case when logs are unavailable because of S3 failure (checking the logs, and other possible actions, checking workspace, and etc)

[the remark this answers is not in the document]
akihikodaki: I'm still not convinced it's prohibitive even if the events are rare. I haven't seen any occasion where a GitHub Actions/GitLab CI task fails with artifact uploads except that the backend is misconfigured or the network is failing, which requires attention.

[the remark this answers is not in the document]
YanVugenfirer: We had several failures:
1. DropBox API issues
2.  Github API issues (failure to update the status of the CI due to GitHub API temporary unavailability)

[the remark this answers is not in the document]
zjmletang: I’d like to express my concern as well. Since we actually support S3-compatible storage, not just limited to AWS and Alibaba Cloud, I'm worried that not all S3-compatible storage solutions are equally stable.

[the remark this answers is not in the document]
akihikodaki: Relaxed rate limiting and high availability are not just of S3 but common for similar services. For example, if such services have strict rate limits, they cannot be used to build Web services. They are intended to be used as building blocks of complex services and should satisfy requirements for such use cases.

lib/resultuploaders/s3/s3.rb
zjmletang: @akihikodaki , Okay, I understand your point. I have two small details I would like to confirm with you:

1.When you mentioned "key," do you mean the full path or just the last part of the file name? I feel that using only the last part of the file name might be more appropriate, as the full path can be quite long.

2.In terms of implementation, I will change the original @uploader_urls to @object_keys = [] to store the key for each object (I won't directly use the objects array because it still needs to handle duplicates based on keys, which may affect performance).

Then the implementation for index.html.erb is as follows:
image

lib/resultuploaders/s3/s3.rb
akihikodaki: Having a method adds a few lines of code. Although the code size increase is quite small, it's better to avoid adding code for future needs; we can always add the code whenever necessary.

lib/resultuploaders/s3/s3.rb
zjmletang: > It would be better to store just the last part of the file names.

@akihikodaki ,The reason I am not using a filename array is that if I do, I would need both urls and filenames arrays simultaneously, and in the ERB, I would need to associate the elements of the two arrays (although this association might be quite simple here). On the other hand, if I use keys or my original urls, I only need one array. What do you think?

akihikodaki: @zjmletang You only need one array either way; create URLs from filenames.

lib/resultuploaders/s3/s3.rb
zjmletang: Okay, I'll make some changes. Thank you for your guidance

lib/resultuploaders/s3/s3.rb
zjmletang: @akihikodaki ，Okay, I misinterpreted your point earlier， I'm not sure if I understood your point now. Could you please confirm: do you mean to pass both @path and @filename into the ERB to construct a key, and then create an object out of that? The URL would still be obtained through public_url, just like in the image I shared above

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,136 @@
...
+        secret_access_key: @secret_access_key,
+        region: @region, endpoint: "https://#{@endpoint}"
+      )
+      @bucket = s3_resource.bucket(@bucket_name)
+
+      @index_template = ERB.new(File.read('lib/templates/index.html.erb'))
+      @index_obj = nil
+      @filenames = []
+    end
+    # rubocop:enable Metrics/AbcSize
```
akihikodaki, line 51: Let's not disable Metrics/AbcSize but split this method.
In particular, you can write a method that returns `Aws::S3::Bucket`. You can convert `@config`, `@access_key_id`, `@secret_access_key`, `@region`, `@bucket_name`, `@endpoint` to local variables of such a method.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      s3_resource = Aws::S3::Resource.new(
+        access_key_id: @access_key_id,
+        secret_access_key: @secret_access_key,
+        region: @region, endpoint: "https://#{@endpoint}"
+      )
+      @bucket = s3_resource.bucket(@bucket_name)
+
+      @index_template = ERB.new(File.read('lib/templates/index.html.erb'))
+      @index_obj = nil
```
akihikodaki, line 48: Initialize `@index_obj` with a real value instead of `nil`.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    def close; end
+
+    private
+
+    def generate_index
+      handle_exceptions(__method__) do
+        index_data = {
+          'title' => @path.split('/').last,
```
akihikodaki, line 126: Perhaps you may just write the entire `@path`. I feel the repository name information is somewhat useful, and it also saves `split` and `last` calls.

akihikodaki (changes requested):

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    def initialize(project)
+      tag = project.engine_tag
+      timestamp = project.timestamp
+      repo = project.config['repository']
+      @path = "#{repo}/CI/#{tag}-#{timestamp}"
+
+      @logger = project.logger
+
+      @bucket = initialize_bucket
```
akihikodaki, line 33: Better to be named as `new_bucket`. In Ruby, `new` creates an object while `initialize` only initializes a created object.

akihikodaki (changes requested):

lib/resultuploaders/s3/s3.rb
zjmletang: Okay, I have made the changes, and I have also addressed the other issues you mentioned.

lib/resultuploaders/s3compatible/s3compatible.rb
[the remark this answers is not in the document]
YanVugenfirer: Again, I must insist on the 100% resilience of the system on our side. 

Any web system has safeguards against 3rd party API failures.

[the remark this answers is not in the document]
akihikodaki: A Web service that cannot store user data is non-functional so the availability of the storage service is critical for resilience. We can say the same for a CI system; a CI system that cannot report the results and store artifacts provides no value and requires intervention to fix the issue.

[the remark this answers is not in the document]
YanVugenfirer: Yes, but the web service that cannot momentarily store your picture will not crash the front end.  And it will not crash the micro service that uses this storage. It will log the error and report the anomaly. 

Crashing hck-ci because we couldn't upload test results (and again, this is not theory) is worse than not crashing HCK-CI and reporting the result as pass or fail. The user that will want to investigate will check what happened. In case of crashing.

Moreover, hck-ci crashes can happen when we upload partial results. Let's say CI ran the first HCK test and wants to upload the test results and update the log file that we have (the status of what tests passed, failed, or pending). It fails. But it might succeed next time. Why would we crash hck-ci in such a case? Even worse - it can be CI run executing for hours, and again, only one test result upload fails. Why would we kill the execution and keep the user with partial results?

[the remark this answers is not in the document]
akihikodaki: > Yes, but the web service that cannot momentarily store your picture will not crash the front end. And it will not crash the micro service that uses this storage. It will log the error and report the anomaly.
> 
> Crashing hck-ci because we couldn't upload test results (and again, this is not theory) is worse than not crashing HCK-CI and reporting the result as pass or fail. The user that will want to investigate will check what happened. In case of crashing.

Such a Web service will still fail for that particular request. This is the same for AutoHCK; one CI run will fail but the other may continue running.

> 
> Moreover, hck-ci crashes can happen when we upload partial results. Let's say CI ran the first HCK test and wants to upload the test results and update the log file that we have (the status of what tests passed, failed, or pending). It fails. But it might succeed next time. Why would we crash hck-ci in such a case? Even worse - it can be CI run executing for hours, and again, only one test result upload fails. Why would we kill the execution and keep the user with partial results?

I'm thinking of this feature to report the results.

> Moreover, hck-ci crashes can happen when we upload partial results. Let's say CI ran the first HCK test and wants to upload the test results and update the log file that we have (the status of what tests passed, failed, or pending). It fails. But it might succeed next time. Why would we crash hck-ci in such a case? Even worse - it can be CI run executing for hours, and again, only one test result upload fails. Why would we kill the execution and keep the user with partial results?

That is the concern of the caller so it's better to be handled there, not at individual uploader implementations as suggested in: https://github.com/HCK-CI/AutoHCK/pull/476#discussion_r1797697163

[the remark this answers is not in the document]
YanVugenfirer: > That is the concern of the caller so it's better to be handled there, not at individual uploader implementations as suggested in: https://github.com/HCK-CI/AutoHCK/pull/476#discussion_r1797697163

So until it is done (let's open an issue), I suggest having exception handling in the individual uploader.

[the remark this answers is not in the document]
akihikodaki: For now, I suggest rescuing `Aws::Errors::ServiceError` and `Seahorse::Client::NetworkingError`. Rescuing `StandardError` will suppress even programming errors like `NoMethodError`. Rescuing these two should suppress errors due to networking and ones reported by the storage service.
We can amend it to raise a special error class instead of logging when we need to handle errors in callers.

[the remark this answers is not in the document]
zjmletang: Okay, I've made the change

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+    # The current exception handling only logs errors without any additional recovery or fallback mechanisms.
+    # It is suggested to rescue only Aws::Errors::ServiceError and Seahorse::Client::NetworkingError,
+    # as rescuing StandardError may suppress other programming errors (e.g., NoMethodError).
+    # This approach focuses on handling networking errors and those reported by the storage service.
+    # Future enhancements may include more robust error handling strategies and raising a special error class
+    # instead of just logging when we need to handle errors in callers.
+    def handle_exceptions(where)
+      yield
+    rescue Aws::Errors::ServiceError, Seahorse::Client::NetworkingError => e
+      @logger.warn("S3 #{where} error: (#{e.class}) #{e.message}")
```
akihikodaki, line 51: Simpler:

```suggestion
      @logger.warn("S3 #{where} error: (#{e.detailed_message}")
```

`e.detailed_message` will give something like `foo (RuntimeError)`.

lib/resultuploaders/s3/s3.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      @index_obj = @bucket.object("#{@path}/#{INDEX_FILE_NAME}")
+      @index_template = ERB.new(File.read('lib/templates/index.html.erb'))
+      @filenames = []
+    end
+
+    def html_url; end
+
+    # The current exception handling only logs errors without any additional recovery or fallback mechanisms.
+    # It is suggested to rescue only Aws::Errors::ServiceError and Seahorse::Client::NetworkingError,
```
akihikodaki, line 43: Let's keep the wording in comments concise.

```suggestion
    # Rescue only Aws::Errors::ServiceError and Seahorse::Client::NetworkingError,
```

akihikodaki (changes requested):
There are few nits remaining and this pull request will be ready once they are handled. Thank you for keeping working on this feature!

akihikodaki (approved):
Now all my feedback is addressed. Thank you for your contribution!

zjmletang:
> There are few nits remaining and this pull request will be ready once they are handled. Thank you for keeping working on this feature!

I have made the changes. Thank you very much for your thorough guidance; I have learned a lot @akihikodaki @kostyanf14 @YanVugenfirer

PR #478, opened 2024-09-25
Support pre/post_test_commands for host/guest

Jedoku:


lib/setupmanagers/hckclient.rb
```diff
@@ -163,7 +164,7 @@ def configure(run_only: false)
         end
 
         configure_machine
-        run_pre_test_commands
+        run_pre_tests_commands
```
kostyanf14, line 167: this part of refactoring should be done in separate commit

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,22 @@ def wait_queued_test(id)
...
+    end
+
+    def test_parameters(test_name)
+      read_test_config(test_name, :parameters)
+        .to_h { |parameter| [parameter.name, parameter.value] }
+    end
+
+    def run_test_commands(test_name, type)
+      read_test_config(test_name, type).each do |commands|
+        @tools.run_on_machine(@client.name, commands.guest, commands.guest) if commands.guest
```
kostyanf14, line 148: a command should be run on all clients (client/support)

lib/models/hcktest_config.rb
```diff
@@ -12,12 +12,21 @@ class Parameter < T::Struct
       const :value, String
     end
 
+    class Command < T::Struct
+      extend T::Sig
+
+      const :guest, T.nilable(String)
```
kostyanf14, line 18: I suggest to add some description and log it before run, to make logs more informative.

lib/models/hcktest_config.rb
```diff
@@ -12,12 +12,21 @@ class Parameter < T::Struct
...
+      const :host, T.nilable(String)
+    end
+
     class TestConfig < T::Struct
       extend T::Sig
 
       const :tests, T::Array[String]
       const :secure, T.nilable(T::Boolean)
       const :parameters, T::Array[Parameter], default: []
+      const :pre_test_commands, T::Array[Command], default: []
```
kostyanf14, line 28: What about reusing existing CommandInfo? Extend it if needed

kostyanf14 (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,22 @@ def wait_queued_test(id)
...
+        spawn(*commands.host, out: :err) if commands.host
```
kostyanf14, line 149: Do these commands run synchronously or asynchronously?
Why do not use the CmdRun class?

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,37 @@ def wait_queued_test(id)
...
-              .to_h { |parameter| [parameter.name, parameter.value] }
+              .flat_map(&config)
+    end
+
+    def test_parameters(test_name)
+      read_test_config(test_name, :parameters)
+        .to_h { |parameter| [parameter.name, parameter.value] }
+    end
+
+    def run_guest_commands(test, commands)
```
akihikodaki, line 146: Here we only give one command for each call so the parameter should be singular.

lib/engines/hcktest/tests.rb
```diff
@@ -433,12 +459,13 @@ def load_clients_system_info
       build_system_info(support_sysinfo)
     end
 
-    def run(tests)
+    def run(tests) # rubocop:disable Metrics/AbcSize
```
akihikodaki, line 462: This method indeed looks cluttered now so let's refactor it. I have two ideas:
- You can move `@total = @tests.count` to `tests_stats` and convert it into a local variable.
- `@last_done = [] if @last_done.nil?` can also be moved to `initialize`. It also makes `if @last_done.nil?` unnecessary.

lib/models/command_info.rb
```diff
@@ -11,7 +11,8 @@ class CommandInfo < T::Struct
       extend JsonHelper
 
       const :desc, String
-      const :run, String
+      const :host_run, T.nilable(String)
```
akihikodaki, line 14: This adds a property but `HCKClient` ignores it. One of the reasons of having a type is to tell the user that there is an extra property and having an extra property in type defeats that purpose.

lib/setupmanagers/hckclient.rb
```diff
@@ -72,11 +72,12 @@ def set_machine_ready
       raise ClientRunError, "Couldn't set #{@name} state to Ready"
     end
 
-    def run_pre_test_commands
+    def run_pre_tests_commands
```
akihikodaki, line 75: - If you rename a method you should do so for the JSON too.
- The differences of "test" and "tests" are too subtle. It is not really related to tests so better avoid using the word "tests". For example, you can name it "post_start".

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,37 @@ def wait_queued_test(id)
       end
     end
 
-    def test_parameters(test_name)
+    def read_test_config(test_name, config)
```
akihikodaki, line 135: "Read" sounds like it's reading a file. An alternative is "select", which is the word used in `Array#select` called in this method.

lib/engines/hcktest/tests.rb
```diff
@@ -447,7 +474,7 @@ def run(tests)
           queue_test(test, wait: true)
           list_tests
           handle_test_running
-
+          run_test_commands(test, :post_test_commands)
```
akihikodaki, line 477: It is weird to run `pre_tests_commands` just once for all runs while running `post_test_commands` for each run.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,37 @@ def wait_queued_test(id)
...
+      @logger.info("Running command (#{commands.desc}) on clients")
+      @tools.run_on_machine(@client.name, commands.desc, commands.guest_run)
+      @tools.run_on_machine(@support.name, commands.desc, commands.guest_run) if support_needed?(test)
+    end
+
+    def run_host_commands(commands)
+      return unless commands.host_run
+
+      @logger.info("Running command (#{commands.desc}) on host")
+      Process.wait spawn(*commands.host_run, out: :err)
```
kostyanf14, line 158: host commands can have replacements as in lib/setupmanagers/qemuhck/devices/tpm-tis.json for example

kostyanf14: Why don't use CmdRun class?

kostyanf14 (changes requested):

lib/engines/hcktest/tests.rb
Jedoku: CmdRun requires scope i just pick easiest way

kostyanf14: But who will control and terminate daemons for example in this case?

lib/setupmanagers/hckclient.rb
```diff
@@ -76,7 +76,7 @@ def run_pre_test_commands
       @project.engine.drivers&.each do |driver|
         driver.pretestcommands&.each do |command|
```
kostyanf14: in this commit you renamed `pretestcommands -> post_start_commands` in JSON but `lib/models/driver.rb` is not updated. Also `def run_pre_test_commands` is not updated. So you have broken AutoHCK after applying this commit.

lib/setupmanagers/hckclient.rb
```diff
@@ -76,10 +78,18 @@ def run_post_start_commands
...
+            @logger.info("Running command (#{desc}) on client #{@name}")
+            @tools.run_on_machine(@name, desc, guest_run)
+          end
 
-          @logger.info("Running command (#{desc}) on client #{@name}")
-          @tools.run_on_machine(@name, desc, cmd)
+          if host_run
+            @logger.info("Running command (#{desc}) on host")
+            run_cmd(host_run)
+          end
```
kostyanf14, line 92: run_post_start_commands is called from HCKClient#configure which is called for each client. So, the run_post_start_commands will be called for each client too, but the host is one and we don't need to execute the host command twice.

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,34 @@ def wait_queued_test(id)
...
+
+    def run_guest_command(test, command)
+      @tools.run_on_machine(@client.name, command, command)
+      @tools.run_on_machine(@support.name, command, command) if support_needed?(test)
+    end
+
+    def run_test_commands(test, type)
+      select_test_config(test['name'], type).each do |commands|
+        if commands.guest_run
+          @logger.info("Running command (#{commands.desc}) on clients")
```
kostyanf14, line 154: Better to have log `@logger.info("Running command (#{commands.desc}) on client #{name}")` instead of a generic one. This is more complicated to debug fail.

kostyanf14 (changes requested):

lib/engines/hcktest/hcktest.rb
```diff
@@ -147,13 +147,25 @@ def run_clients(scope, run_opts = {})
...
+    end
+
     def configure_and_synchronize_clients
       run_only = @project.options.test.manual && @project.options.test.driver_path.nil?
 
       @clients.each_value do |client|
         client.configure(run_only:)
       end
 
+      run_clients_post_start_host_commands
```
akihikodaki, line 168: Why is this change in `HCKTest` while `guest_run` is handled in `HCKClient`?

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,36 @@ def wait_queued_test(id)
...
+    end
+
+    def run_guest_command(test, command, desc)
+      @logger.info("Running command (#{desc}) on client #{@client.name}")
+      @tools.run_on_machine(@client.name, command, command)
+
+      return unless support_needed?(test)
+
+      @logger.info("Running command (#{desc}) on client #{@support.name}")
+      @tools.run_on_machine(@support.name, command, command)
```
akihikodaki, line 153: This method has code duplication for `@client` and `@support`. It is better to have a method that runs a command on one client instead of this to reduce redundancy.

akihikodaki (changes requested):

lib/engines/hcktest/hcktest.rb
kostyanf14: Because of this comment  https://github.com/HCK-CI/AutoHCK/pull/478#discussion_r1785852107

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,34 @@ def wait_queued_test(id)
...
+
+    def run_command_on_client(client, command, desc)
+      @logger.info("Running command (#{desc}) on client #{client.name}")
+      @tools.run_on_machine(client.name, command, command)
+    end
+
+    def run_test_commands(test, type)
+      select_test_config(test['name'], type).each do |commands|
+        if commands.guest_run
+          run_command_on_client(@client, command, desc)
```
akihikodaki, line 154: `command` is not defined.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,36 @@ def wait_queued_test(id)
...
+        .to_h { |parameter| [parameter.name, parameter.value] }
+    end
+
+    def run_command_on_client(client, command, desc)
+      @logger.info("Running command (#{desc}) on client #{client.name}")
+      @tools.run_on_machine(client.name, desc, command)
+    end
+
+    def run_test_commands(test, type)
+      select_test_config(test['name'], type).each do |commands|
```
akihikodaki, line 152: You name variables holding `CommandInfo` as "command" instead of "commands". Keep them consistent.

lib/engines/hcktest/tests.rb
```diff
@@ -132,11 +132,36 @@ def wait_queued_test(id)
...
+        guest_cmd = commands.guest_run
+        description = commands.desc
```
akihikodaki, line 154: Let's name it `desc` for consistency too.

akihikodaki (changes requested):

akihikodaki (approved):

lib/engines/hcktest/drivers/NetKVM.json
```diff
@@ -5,7 +5,7 @@
   "install_method": "PNP",
```
kostyanf14, line 5: Don't use a long commit message title. The good practice to have the commit message title no longer than 72 characters. If this is needed you can add more information in the description.

kostyanf14 (approved):

PR #486, opened 2024-10-03
Flush stor

kostyanf14:


bin/fake-snmp-reset
```diff
@@ -0,0 +1,71 @@
...
+    @qmp_socket.write JSON.dump({ 'execute' => cmd })
+    @qmp_socket.flush
+
+    loop do
+      response = JSON.parse(@qmp_socket.readline)
+      break response['return'] if response.key?('return')
+      raise response['error'].to_s if response.key?('error')
+    end
+  end
+end
```
akihikodaki, line 64: I prefer having stateless methods instead of having a class with full of states. Especially we can connect QMP each time to make the code simpler.

bin/fake-snmp-reset
```diff
@@ -0,0 +1,71 @@
...
+
+  def send_cmd(cmd)
+    @logger.info "Sending #{cmd} via QMP"
+    @qmp_socket.write JSON.dump({ 'execute' => cmd })
+    @qmp_socket.flush
+
+    loop do
+      response = JSON.parse(@qmp_socket.readline)
+      break response['return'] if response.key?('return')
+      raise response['error'].to_s if response.key?('error')
```
akihikodaki, line 61: Use `@logger.error` and break to keep the server alive in case of error.

bin/fake-snmp-reset
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+end
+
+log = Logger.new('fake-snmp-reset.txt')
+log.level = Logger::DEBUG
```
akihikodaki, line 67: This is unnecessary; `DEBUG` is the default log level.

bin/fake-snmp-reset
```diff
@@ -0,0 +1,71 @@
...
+log.info "START #{ARGV}"
```
akihikodaki, line 68: Let's remove this; I don't see a point of failure between this log and "Initializing UDP socket".

bin/ns_unshared
```diff
@@ -40,6 +40,10 @@ module AutoHCK
       %w[ip link set br_world up],
       %w[ip route add default via 10.0.2.2],
       %w[ip link add br_ctrl type bridge],
+      # See HLK-Setup-Scripts IP configuration
+      # https://github.com/HCK-CI/HLK-Setup-Scripts/blob/master/client.ps1
+      # need for access to host without full Internet access
```
akihikodaki, line 45: nit:

```suggestion
      # needed for access to host without full Internet access
```

akihikodaki (changes requested):

akihikodaki (approved):

YanVugenfirer (approved):

bin/fake-snmp-reset
```diff
@@ -0,0 +1,50 @@
...
+
+    @logger.error(response['error'].to_s) if response.key?('error')
+  end
+end
+
+count = 0
+
+loop do
+  @logger.info 'Initializing UDP socket'
+  UDPSocket.open do |socket|
```
akihikodaki, line 32: I think you can put `UDPSocket.open` before the loop now.

akihikodaki (changes requested):
https://oidref.com/1.3.6.1.4.1.318.1.1.12.3.3.1.1.4 says it has `delayedReboot` so you may check if the value is used. I don't require to do so as Microsoft's documentation does not mention it.

bin/fake-snmp-reset
```diff
@@ -0,0 +1,50 @@
...
+    socket.bind('0.0.0.0', 'snmp')
+    @logger.info "Listening on #{socket.local_address.inspect_sockaddr}"
+
+    _, addr = socket.recvfrom(512)
```
akihikodaki, line 36: To discard all data:

```suggestion
    _, addr = socket.recvfrom(0)
```

akihikodaki (approved):

YanVugenfirer (approved):

benyamin-codez:
@kostyanf14

Thanks for the reference.

I noticed the [MS doco](https://learn.microsoft.com/en-us/windows-hardware/test/hlk/testref/2dec1f67-b506-4434-8bdf-763147ad8f0b#power-strip) suggested the following ASNs are required:

* ImmediatePowerOn = 1
* ImmediatePowerOff = 2
* ImmediateReboot = 3

_[sic] - I'm presuming typographic errors here, and these should actually be:_

* immediateOn = 1
* immediateOff = 2
* immediateReboot = 3

...from https://oidref.com/1.3.6.1.4.1.318.1.1.12.3.3.1.1.4 (the relevant OID):

```
rPDUOutletControlOutletCommand OBJECT-TYPE
   SYNTAX INTEGER {
      immediateOn             (1),
      immediateOff            (2),
      immediateReboot         (3),
      delayedOn               (4),
      delayedOff              (5),
      delayedReboot           (6),
      cancelPendingCommand    (7)
   }
```

As the first two, i.e. _immediateOn_ and _immediateOff_, are `read-write`, i.e. `set/get`, perhaps we are performing the `system-reset` via QMP, not only when the test issues the command but perhaps also when the test issues a query to check the power state.

***So when asked to check the power state, perhaps we are resetting instead...*** 8^O

This might explain the erratic results seen in virtio-win...
Perhaps we need to tee the socket data to a SNMP parser to check...
...but it might be quicker / easier to add the requisite features to the listener here... 8^d

kostyanf14:
@benyamin-codez 

No, you are not correct. We analyze requests, and HLK always sends `ImmediateReboot` and nothing else.

kostyanf14:
@benyamin-codez Also, if you open HLKX results where Flush test failed, you can see the error. Error looks like this: System reset was requested but system not rebooted.

benyamin-codez:
@kostyanf14 

Thanks for the prompt reply.

> No, you are not correct. We analyze requests, and HLK always sends ImmediateReboot and nothing else.

I saw the `socket.recvfrom(0)`, with the zero-sized buffer, and presumed you weren't looking...

> Also, if you open HLKX results where Flush test failed, you can see the error. Error looks like this:
> System reset was requested but system not rebooted.

Oh, ok. I was checking `Failed: Flush Test.zip` > `Test Run` > `WttEA.log` instead
I've only seen `0x80070002` (`ERROR_FILE_NOT_FOUND` errors) in the dozen or so I've checked.

I'll grab the Studio and have a peek... 8^d
Can you tell I'm trying to avoid installing the whole HLK...?  8^D  It seems like quite a beast...

kostyanf14:
> I saw the `socket.recvfrom(0)`, with the zero-sized buffer, and presumed you weren't looking...

During development, we checked this data and found that it can be ignored. 

For example:
https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1304 - viostor-Win2022x64
image
 
The error is not related to the storage driver but something with HLK.

benyamin-codez:
Yeah, that's what I see in the few I peeked into.
The actual test passes and then an error occurs during test cleanup.
Apparently it occurs whilst _"Initializing and formatting the disk"_...
So it doesn't ***appear*** to be related to this SNMP kit or reboots at all - at least in these cases.

Is it possible the disk is OFFLINE and so it cannot be initialised or formatted?
... If so, this is something we might need to address in the driver...
Is it perhaps something to do with the type of disk backing, i.e. COW or RAW?
This appears to be running (and failing) on non-bootable disks (as expected - except the failing part)...
I presume there's no AV or anything else interfering with anything in the boot block...

Quite annoying...! 8^(

kostyanf14:
You can configure only the following options:
- IP 	IP Address of Remote PDU
- OID 	OID of Remote PDU outlet
- Outlet 	Port of Remote PDU outlet
- Community 	Community of Remote PDU (e.g. private)

All other configurations are done automatically by HLK (not AutoHCK)

kostyanf14:
@benyamin-codez Let's continue in the discussion thread https://github.com/HCK-CI/AutoHCK/discussions/608. This is more complicated to find this PR and this creates extra mail noise for other people.

PR #487, opened 2024-10-07
RHELMISC-4745: Add ability to bring up previous session

Jedoku:
WIP

### DIFF
--- a/bin/auto_hck
+++ b/bin/auto_hck
@@ -9,6 +9,13 @@ module AutoHCK
   run do
     cli = CLI.new
     cli.parse(ARGV)
+    session_path = cli.test.latest_session ? "#{Config.read['workspace_path']}/latest" : cli.test.session
+
+    session = if session_path
+                Session.load(session_path, cli.common.workspace_path)
+              else
+                Session.new(cli: cli)
+              end
 
     ENV.store 'LC_ALL', 'en_US.UTF-8'
 
@@ -18,7 +25,7 @@ module AutoHCK
     Thread.report_on_exception = false
 
     ResourceScope.open do |scope|
-      @project = Project.new(scope, cli)
+      @project = Project.new(scope, session)
       Trap.project = @project
       @project.run if @project.prepare
     rescue StandardError => e
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -1,252 +1,278 @@
+# typed: true
 # frozen_string_literal: true
 
 # AutoHCK module
 module AutoHCK
-  # class CLI
-  class CLI
-    attr_reader :common, :install, :test, :mode
+  # class CommonOptions
+  class CommonOptions < T::Struct
+    include T::Sig
 
-    def initialize
-      @common = CommonOptions.new
-      @test = TestOptions.new
-      @install = InstallOptions.new
+    prop :verbose, T::Boolean, default: false
+    prop :config, T.nilable(String)
+    prop :client_world_net, T::Boolean, default: false
+    prop :id, Integer, default: 2
+    prop :share_on_host_path, T.nilable(String)
+    prop :workspace_path, T.nilable(String)
+    prop :client_ctrl_net_dev, T.nilable(String)
 
-      @sub_parser = {
-        'test' => @test.create_parser,
-        'install' => @install.create_parser
-      }
-
-      @parser = @common.create_parser(@sub_parser)
-    end
-
-    # class CommonOptions
-    class CommonOptions
-      attr_accessor :verbose, :config, :client_world_net, :id, :share_on_host_path, :workspace_path,
-                    :client_ctrl_net_dev
-
-      def create_parser(sub_parser)
-        OptionParser.new do |parser|
-          parser.banner = 'Usage: auto_hck.rb [common options] <command> [command options]'
-          parser.separator ''
-          define_options(parser)
-          parser.on_tail('-h', '--help', 'Show this message') do
-            puts parser
-            sub_parser&.each_value do |v|
-              puts v
-            end
-            exit
+    def create_parser(sub_parser)
+      OptionParser.new do |parser|
+        parser.banner = 'Usage: auto_hck.rb [common options] <command> [command options]'
+        parser.separator ''
+        define_options(parser)
+        parser.on_tail('-h', '--help', 'Show this message') do
+          puts parser
+          sub_parser&.each_value do |v|
+            puts v
           end
+          exit
         end
       end
+    end
 
-      # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-      def define_options(parser)
-        @verbose = false
-        @config = nil
-        @client_world_net = false
-        @id = 2
-        @share_on_host_path = nil
+    # rubocop:disable Metrics/MethodLength
+    def define_options(parser)
+      parser.on('--share-on-host-path <path>', String,
+                'For using Transfer Network specify the directory to share on host machine') do |share_on_host_path|
+        @share_on_host_path = share_on_host_path
+      end
 
-        parser.on('--share-on-host-path <path>', String,
-                  'For using Transfer Network specify the directory to share on host machine') do |share_on_host_path|
-          @share_on_host_path = share_on_host_path
+      parser.on('--verbose', TrueClass,
+                'Enable verbose logging',
+                &method(:verbose=))
+
+      parser.on('--config <override.json>', String,
+                'Path to custom override.json file',
+                &method(:config=))
+
+      parser.on('--client_world_net', TrueClass,
+                'Attach world bridge to clients VM',
+                &method(:client_world_net=))
+
+      parser.on('--client-ctrl-net-dev <client-ctrl-net-dev>', String,
+                'Client VM control network device (make sure that driver is installed)',
+                &method(:client_ctrl_net_dev=))
+
+      parser.on('--id <id>', Integer,
+                'Set ID for AutoHCK run',
+                &method(:id=))
+
+      parser.on('-v', '--version',
+                'Display version information and exit') do
+        puts "AutoHCK Version: #{AutoHCK::VERSION}"
+        exit
+      end
+
+      parser.on('-w <path>', String,
+                'Internal use only',
+                &method(:workspace_path=))
+    end
+  end
+
+  # class TestOptions
+  class TestOptions < T::Struct
+    include T::Sig
+
+    prop :platform, T.nilable(String)
+    prop :drivers, T::Array[String], default: []
+    prop :driver_path, T.nilable(String)
+    prop :commit, T.nilable(String)
+    prop :svvp, T::Boolean, default: false
+    prop :dump, T::Boolean, default: false
+    prop :gthb_context_prefix, T.nilable(String)
+    prop :gthb_context_suffix, T.nilable(String)
+    prop :playlist, T.nilable(String)
+    prop :select_test_names, T.nilable(String)
+    prop :reject_test_names, T.nilable(String)
+    prop :reject_report_sections, T::Array[String], default: []
+    prop :boot_device, T.nilable(String)
+    prop :allow_test_duplication, T::Boolean, default: false
+    prop :manual, T::Boolean, default: false
+    prop :package_with_playlist, T::Boolean, default: false
+    prop :enable_vbs, T::Boolean, default: false
+    prop :session, T.nilable(String)
+    prop :latest_session, T::Boolean, default: false
+
+    def create_parser
+      OptionParser.new do |parser|
+        parser.banner = 'Usage: auto_hck.rb test [test options]'
+        parser.separator ''
+        define_options(parser)
+        parser.on_tail('-h', '--help', 'Show this message') do
+          puts parser
+          exit
         end
+      end
+    end
 
-        parser.on('--verbose', TrueClass,
-                  'Enable verbose logging',
-                  &method(:verbose=))
+    def define_options(parser) # rubocop:disable Metrics/AbcSize
+      @reject_report_sections = []
 
-        parser.on('--config <override.json>', String,
-                  'Path to custom override.json file',
-                  &method(:config=))
+      parser.on('-p', '--platform <platform_name>', String,
+                'Platform for run test',
+                &method(:platform=))
 
-        parser.on('--client_world_net', TrueClass,
-                  'Attach world bridge to clients VM',
-                  &method(:client_world_net=))
+      parser.on('-d', '--drivers <drivers_list>', Array,
+                'List of driver for run test',
+                &method(:drivers=))
 
-        parser.on('--client-ctrl-net-dev <client-ctrl-net-dev>', String,
-                  'Client VM control network device (make sure that driver is installed)',
-                  &method(:client_ctrl_net_dev=))
+      parser.on('--driver-path <driver_path>', String,
+                'Path to the location of the driver wanted to be tested',
+                &method(:driver_path=))
 
-        parser.on('--id <id>', Integer,
-                  'Set ID for AutoHCK run',
-                  &method(:id=))
+      parser.on('-c', '--commit <commit_hash>', String,
+                'Commit hash for CI status update',
+                &method(:commit=))
 
-        parser.on('-v', '--version',
-                  'Display version information and exit') do
-          puts "AutoHCK Version: #{AutoHCK::VERSION}"
+      parser.on('--svvp', TrueClass,
+                'Run SVVP tests for specified platform instead of driver tests',
+                &method(:svvp=))
+
+      parser.on('--dump', TrueClass,
+                'Create machines snapshots and generate scripts for run it manualy',
+                &method(:dump=))
+
+      parser.on('--gthb_context_prefix <gthb_context_prefix>', String,
+                'Add custom prefix for GitHub CI results context',
+                &method(:gthb_context_prefix=))
+
+      parser.on('--gthb_context_suffix <gthb_context_suffix>', String,
+                'Add custom suffix for GitHub CI results context',
+                &method(:gthb_context_suffix=))
+
+      parser.on('--playlist <playlist>', String,
+                'Use custom Microsoft XML playlist',
+                &method(:playlist=))
+
+      parser.on('--select-test-names <select_test_names>', String,
+                'Use custom user text playlist',
+                &method(:select_test_names=))
+
+      parser.on('--reject-test-names <reject_test_names>', String,
+                'Use custom CI text ignore list',
+                &method(:reject_test_names=))
+
+      parser.on('--reject-report-sections <reject_report_sections>', Array,
+                'List of section to reject from HTML results',
+                '(use "--reject-report-sections=help" to list sections)') do |reject_report_sections|
+        if reject_report_sections.first == 'help'
+          puts Tests::RESULTS_REPORT_SECTIONS.join("\n")
           exit
         end
 
-        parser.on('-w <path>', String,
-                  'Internal use only',
-                  &method(:workspace_path=))
+        extra_keys = reject_report_sections - Tests::RESULTS_REPORT_SECTIONS
+
+        raise(AutoHCKError, "Unknown report sections: #{extra_keys.join(', ')}.") unless extra_keys.empty?
+
+        @reject_report_sections = reject_report_sections
       end
-      # rubocop:enable Metrics/AbcSize,Metrics/MethodLength
+
+      parser.on('--boot-device <boot_device>', String,
+                'VM boot device',
+                &method(:boot_device=))
+
+      parser.on('--allow-test-duplication', TrueClass,
+                'Allow run the same test several times.',
+                'Works only with custom user text playlist.',
+                'Test results table can be broken. (experimental)',
+                &method(:allow_test_duplication=))
+
+      parser.on('--manual', TrueClass,
+                'Run AutoHCK in manual mode',
+                &method(:manual=))
+
+      parser.on('--package-with-playlist', TrueClass,
+                'Load playlist into HLKX project package',
+                &method(:package_with_playlist=))
+
+      parser.on('--session <path>', String,
+                'Load session from workspace',
+                &method(:session=))
+
+      parser.on('--latest-session', TrueClass,
+                'Load previous session',
+                &method(:latest_session=))
     end
+    # rubocop:enable Metrics/MethodLength
+  end
 
-    # class TestOptions
-    class TestOptions
-      attr_accessor :platform, :drivers, :driver_path, :commit, :svvp, :dump,
-                    :gthb_context_prefix, :gthb_context_suffix, :playlist, :select_test_names,
-                    :reject_test_names, :reject_report_sections, :boot_device,
-                    :allow_test_duplication, :manual, :package_with_playlist, :enable_vbs
+  # class InstallOptions
+  class InstallOptions < T::Struct
+    include T::Sig
 
-      def create_parser
-        OptionParser.new do |parser|
-          parser.banner = 'Usage: auto_hck.rb test [test options]'
-          parser.separator ''
-          define_options(parser)
-          parser.on_tail('-h', '--help', 'Show this message') do
-            puts parser
-            exit
-          end
+    prop :platform, T.nilable(String)
+    prop :force, T::Boolean, default: false
+    prop :skip_client, T::Boolean, default: false
+    prop :drivers, T::Array[String], default: []
+    prop :driver_path, T.nilable(String)
+    prop :debug, T::Boolean, default: false
+
+    def create_parser
+      OptionParser.new do |parser|
+        parser.banner = 'Usage: auto_hck.rb install [install options]'
+        parser.separator ''
+        define_options(parser)
+        parser.on_tail('-h', '--help', 'Show this message') do
+          puts parser
+          exit
         end
       end
-
-      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-      def define_options(parser)
-        @reject_report_sections = []
-        @svvp = false
-        @enable_vbs = false
-
-        parser.on('-p', '--platform <platform_name>', String,
-                  'Platform for run test',
-                  &method(:platform=))
-
-        parser.on('-d', '--drivers <drivers_list>', Array,
-                  'List of driver for run test',
-                  &method(:drivers=))
-
-        parser.on('--driver-path <driver_path>', String,
-                  'Path to the location of the driver wanted to be tested',
-                  &method(:driver_path=))
-
-        parser.on('-c', '--commit <commit_hash>', String,
-                  'Commit hash for CI status update',
-                  &method(:commit=))
-
-        parser.on('--svvp', TrueClass,
-                  'Run SVVP tests for specified platform instead of driver tests',
-                  &method(:svvp=))
-
-        parser.on('--dump', TrueClass,
-                  'Create machines snapshots and generate scripts for run it manualy',
-                  &method(:dump=))
-
-        parser.on('--gthb_context_prefix <gthb_context_prefix>', String,
-                  'Add custom prefix for GitHub CI results context',
-                  &method(:gthb_context_prefix=))
-
-        parser.on('--gthb_context_suffix <gthb_context_suffix>', String,
-                  'Add custom suffix for GitHub CI results context',
-                  &method(:gthb_context_suffix=))
-
-        parser.on('--playlist <playlist>', String,
-                  'Use custom Microsoft XML playlist',
-                  &method(:playlist=))
-
-        parser.on('--select-test-names <select_test_names>', String,
-                  'Use custom user text playlist',
-                  &method(:select_test_names=))
-
-        parser.on('--reject-test-names <reject_test_names>', String,
-                  'Use custom CI text ignore list',
-                  &method(:reject_test_names=))
-
-        parser.on('--enable-vbs', TrueClass,
-                  'Enable VBS state for clients',
-                  &method(:enable_vbs=))
-
-        parser.on('--reject-report-sections <reject_report_sections>', Array,
-                  'List of section to reject from HTML results',
-                  '(use "--reject-report-sections=help" to list sections)') do |reject_report_sections|
-          if reject_report_sections.first == 'help'
-            puts Tests::RESULTS_REPORT_SECTIONS.join("\n")
-            exit
-          end
-
-          extra_keys = reject_report_sections - Tests::RESULTS_REPORT_SECTIONS
-
-          raise(AutoHCKError, "Unknown report sections: #{extra_keys.join(', ')}.") unless extra_keys.empty?
-
-          @reject_report_sections = reject_report_sections
-        end
-
-        parser.on('--boot-device <boot_device>', String,
-                  'VM boot device',
-                  &method(:boot_device=))
-
-        parser.on('--allow-test-duplication', TrueClass,
-                  'Allow run the same test several times.',
-                  'Works only with custom user text playlist.',
-                  'Test results table can be broken. (experimental)',
-                  &method(:allow_test_duplication=))
-
-        parser.on('--manual', TrueClass,
-                  'Run AutoHCK in manual mode',
-                  &method(:manual=))
-
-        parser.on('--package-with-playlist', TrueClass,
-                  'Load playlist into HLKX project package',
-                  &method(:package_with_playlist=))
-      end
-      # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
     end
 
-    # class InstallOptions
-    class InstallOptions
-      attr_accessor :platform, :force, :skip_client, :drivers, :driver_path, :debug
+    def define_options(parser)
+      parser.on('--debug', TrueClass, 'Enable debug mode',
+                &method(:debug=))
 
-      def create_parser
-        OptionParser.new do |parser|
-          parser.banner = 'Usage: auto_hck.rb install [install options]'
-          parser.separator ''
-          define_options(parser)
-          parser.on_tail('-h', '--help', 'Show this message') do
-            puts parser
-            exit
-          end
-        end
-      end
+      parser.on('-p', '--platform <platform_name>', String,
+                'Install VM for specified platform',
+                &method(:platform=))
 
-      # rubocop:disable Metrics/MethodLength
-      def define_options(parser)
-        @force = false
-        @skip_client = false
-        @drivers = []
-        @debug = false
+      parser.on('-f', '--force', TrueClass,
+                'Install all VM, replace studio if exist',
+                &method(:force=))
 
-        parser.on('--debug', TrueClass, 'Enable debug mode',
-                  &method(:debug=))
+      parser.on('--skip_client', TrueClass,
+                'Skip client images installation',
+                &method(:skip_client=))
 
-        parser.on('-p', '--platform <platform_name>', String,
-                  'Install VM for specified platform',
-                  &method(:platform=))
+      parser.on('-d', '--drivers <drivers_list>', Array,
+                'List of driver attach in install',
+                &method(:drivers=))
 
-        parser.on('-f', '--force', TrueClass,
-                  'Install all VM, replace studio if exist',
-                  &method(:force=))
+      parser.on('--driver-path <driver_path>', String,
+                'Path to the location of the driver wanted to be installed',
+                &method(:driver_path=))
+    end
+  end
 
-        parser.on('--skip_client', TrueClass,
-                  'Skip client images installation',
-                  &method(:skip_client=))
+  # class CLI
+  class CLI < T::Struct
+    extend AutoHCK::Models::JsonHelper
+    extend T::Sig
 
-        parser.on('-d', '--drivers <drivers_list>', Array,
-                  'List of driver attach in install',
-                  &method(:drivers=))
+    prop :test, TestOptions, factory: -> { TestOptions.new }
+    prop :common, CommonOptions, factory: -> { CommonOptions.new }
+    prop :install, InstallOptions, factory: -> { InstallOptions.new }
+    prop :mode, T.nilable(String), default: nil
 
-        parser.on('--driver-path <driver_path>', String,
-                  'Path to the location of the driver wanted to be installed',
-                  &method(:driver_path=))
-      end
-      # rubocop:enable Metrics/MethodLength
+    sig { returns(T::Hash[String, OptionParser]) }
+    def sub_parser
+      {
+        'test' => test.create_parser,
+        'install' => install.create_parser
+      }
     end
 
+    sig { returns(OptionParser) }
+    def parser
+      common.create_parser(sub_parser)
+    end
+
+    sig { params(args: T::Array[String]).returns(T::Array[String]) }
     def parse(args)
-      left = @parser.order(args)
+      left = parser.order(args)
       @mode = left.shift
-      @sub_parser[@mode]&.order!(left) unless @mode.nil?
+      sub_parser[@mode]&.order!(left) unless @mode.nil?
     end
   end
 end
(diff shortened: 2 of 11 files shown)


akihikodaki:
I'm not a fan of this idea. It brings significant complexity similar to `--dump`, which adds an alternative way to run tests.
Its use case is also unclear; I often repeatedly run tests using the command line history and I'm satisfied with it.

lib/project.rb
```diff
@@ -157,31 +157,50 @@ def check_run_termination
...
+    def add_latest_alias
       begin
         File.delete("#{@config['workspace_path']}/latest")
       rescue Errno::ENOENT
         # firts run, no symlink to delete
       end
 
       File.symlink(@workspace_path, "#{@config['workspace_path']}/latest")
-      @setup_manager_type&.enter @workspace_path
     end
```
akihikodaki, line 204: `@options.test.latest_session` is ignored.

lib/project.rb
```diff
@@ -157,31 +157,50 @@ def check_run_termination
...
+        @restored = @options.test.session
+        return
+      end
+
+      @options.test.session ? session : create_session
+      add_latest_alias if @options.test.session.nil?
+      @setup_manager_type&.enter @workspace_path
+    end
+
+    def add_latest_alias
```
akihikodaki, line 196: Use "symlink" instead of introducing another word "alias".

lib/project.rb
```diff
@@ -157,31 +157,50 @@ def check_run_termination
...
         FileUtils.mkdir_p(@workspace_path)
       rescue Errno::EEXIST
         @logger.warn('Workspace path already exists')
       end
       @logger.info("Workspace path: #{@workspace_path}")
 
+      Session.save(@workspace_path, @options)
+    end
+
+    def init_session
```
akihikodaki, line 184: This method should still be named `init_workspace`. It initialized more than `session.json`, which is part of workspace.

akihikodaki (changes requested):

lib/project.rb
```diff
@@ -145,31 +145,50 @@ def check_run_termination
...
 
+      @logger.info("Loading session from #{@workspace_path}")
+    end
+
+    def create_session
       @workspace_path = File.join(@config['workspace_path'],
                                   @engine_name,
                                   @engine_tag,
                                   @timestamp)
+
```
akihikodaki, line 161: Avoid an irrelevant change here.

lib/session.rb
```diff
@@ -0,0 +1,76 @@
...
+          'package_with_playlist' => options.test.package_with_playlist
+        },
+        'common' => {
+          'verbose' => options.common.verbose,
+          'config' => options.common.config,
+          'client_world_net' => options.common.client_world_net,
+          'id' => options.common.id,
+          'share_on_host_path' => options.common.share_on_host_path
+        }
+      }.to_json
```
akihikodaki, line 73: Use `T::Struct#serialize` here.

lib/models/session.rb
```diff
@@ -0,0 +1,43 @@
...
+        const :verbose, T.nilable(T::Boolean)
+        const :config, T.nilable(String)
+        const :client_world_net, T.nilable(T::Boolean)
+        const :id, T.nilable(String)
+        const :share_on_host_path, T.nilable(String)
+      end
+
+      const :test, Test
+      const :common, T.nilable(Common)
+    end
```
akihikodaki, line 41: This duplicates the definitions of attributes of `AutoHCK::Project`. Unify them.

akihikodaki (changes requested):

akihikodaki (changes requested):

lib/session.rb
```diff
@@ -0,0 +1,53 @@
...
+      cli.test.allow_test_duplication ||= session.test.allow_test_duplication
+      cli.test.manual ||= true
+      cli.test.package_with_playlist |= session.test.package_with_playlist
+      cli.test.session = session_path(cli)
+      cli.test.latest_session ||= session.test.latest_session
+      cli.common.verbose ||= session.common.verbose
+      cli.common.config ||= session.common.config
+      cli.common.client_world_net ||= session.common.client_world_net
+      cli.common.id ||= session.common.id
+      cli.common.share_on_host_path ||= session.common.share_on_host_path
```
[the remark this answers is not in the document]
akihikodaki: Forget what I said about `clone` on the chat; it makes no difference at all.
I think the reason why it's looping when assigning `session.test` is that it restores the value of the `workspace_path` property, which was previously ignored. Please check what value it should have.

lib/session.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def self.session_path(cli)
+      cli.test.latest_session ? "#{Config.read['workspace_path']}/latest" : cli.test.session
+    end
+
+    private_class_method def self.compose_session_json(options)
+      {
+        test: options.test.serialize,
+        common: options.common.serialize
+      }.to_json
```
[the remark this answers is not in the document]
akihikodaki: We should use `Models::Session` because it's type-checked. If you directly create a hash, it will not be type-checked so is more error-prone.

lib/cli.rb
```diff
@@ -20,8 +21,15 @@ def initialize
     end
 
     # class CommonOptions
-    class CommonOptions
-      attr_accessor :verbose, :config, :client_world_net, :id, :share_on_host_path, :workspace_path
+    class CommonOptions < T::Struct
+      include T::Sig
+
+      prop :verbose, T.nilable(T::Boolean), default: false
```
akihikodaki, line 27:
```suggestion
      prop :verbose, T::Boolean, default: false
```

lib/cli.rb
```diff
@@ -179,13 +197,28 @@ def define_options(parser)
...
-      attr_accessor :platform, :force, :skip_client, :drivers, :driver_path, :debug
+    class InstallOptions < T::Struct
+      include T::Sig
+
+      prop :platform, T.nilable(String)
+      prop :force, T::Boolean, default: false
+      prop :skip_client, T::Boolean, default: false
+      prop :drivers, T::Array[String], default: []
+      prop :driver_path, T.nilable(String)
+      prop :debug, T::Boolean, default: false
```
akihikodaki, line 221: Remove the default value assignments in `define_options` as you did for `CommonOptions`.

lib/session.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def self.load(cli)
+      session = Models::Session.from_json_file("#{session_path(cli)}/session.json")
+
+      backup_cli = cli.clone
+      cli.test = session.test
+      cli.common = session.common
+
+      cli.test.manual ||= true
```
akihikodaki, line 21: Simpler:

```suggestion
      cli.test.manual = true
```

lib/models/session.rb
```diff
@@ -0,0 +1,22 @@
...
+# frozen_string_literal: true
+
+module AutoHCK
+  module Models
+    class Session < T::Struct
+      extend T::Sig
+      extend JsonHelper
+
+      const :test, AutoHCK::CLI::TestOptions
+      const :common, T.nilable(AutoHCK::CLI::CommonOptions)
```
akihikodaki, line 11: Why will `common` be nil?

lib/engines/hcktest/hcktest.rb
```diff
@@ -201,7 +201,7 @@ def upload_driver_package
 
       r_name = "#{@project.engine_tag}.zip"
       zip_path = "#{@project.workspace_path}/#{r_name}"
-      create_zip_from_directory(zip_path, @driver_path)
+      create_zip_from_directory(zip_path, @driver_path) unless File.exist?(zip_path)
```
akihikodaki, line 204: Can't you skip the whole `upload_driver_package` method?

lib/engines/hcktest/hcktest.rb
```diff
@@ -274,7 +280,7 @@ def run_tests_with_config
         next if tests.empty?
 
         ResourceScope.open do |scope|
-          run_clients_and_configure_setup(scope, group => true, create_snapshot: false, boot_from_snapshot: true)
+          run_clients_and_configure_setup(scope, group => true)
```
akihikodaki, line 283: This change will create a snapshot for every group when not restored.

lib/project.rb
```diff
@@ -162,14 +165,30 @@ def init_workspace
...
+    end
+
+    def init_workspace
+      @restored = @options.test.session
+      unless @options.common.workspace_path.nil?
+        @workspace_path = @options.common.workspace_path
+        return
+      end
+
+      @options.test.session ? session : create_session
```
akihikodaki, line 178: `session` is a confusing name; it is a noun but has side effects.

lib/session.rb
```diff
@@ -0,0 +1,41 @@
...
+    include Helper
+
+    def self.save(workspace_path, options)
+      File.write("#{workspace_path}/session.json", compose_session_json(options, workspace_path))
+    end
+
+    def self.load(cli)
+      session = Models::Session.from_json_file("#{session_path(cli)}/session.json")
+
+      backup_cli = cli.clone
```
akihikodaki, line 17: This `clone` can be removed by modifying `session` according to `cli` and then overwriting `cli`.

lib/models/session.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      sig { returns(String) }
+      def as_json
+        {
+          test: test.serialize,
+          common: common&.serialize
+        }.to_json
```
akihikodaki, line 18: Use `Session#serialize`.

lib/project.rb
```diff
@@ -145,12 +145,15 @@ def check_run_termination
...
-      end
+    def session
+      @workspace_path = @options.test.session
+
+      raise AutoHCKError, 'Workspace path does not exist could not load session' unless File.directory?(@workspace_path)
 
+      @logger.info("Loading session from #{@workspace_path}")
+    end
+
+    def create_session
```
akihikodaki, line 156: It says `create_session` but it also creates the workspace directory, which is confusing.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -329,7 +330,16 @@ def options_replacement_map
...
+      filter_commands(@device_infos.map(&:config_commands).flatten.compact)
+    end
+
+    def filter_commands(commands)
+      commands.reject do |cmd|
+        if cmd.include?('@swtpm_setup_bin@') && @run_opts[:reuse_tpm]
+          @logger.warn('Skipping SWTPM installation')
+          true
+        end
+      end
```
kostyanf14, line 342: I don't agree with filters in general. When we bring up the previous session all config commands were executed, so we should skip all of them.

lib/engines/hcktest/hcktest.rb
```diff
@@ -274,7 +281,7 @@ def run_tests_with_config
         next if tests.empty?
 
         ResourceScope.open do |scope|
-          run_clients_and_configure_setup(scope, group => true, create_snapshot: false, boot_from_snapshot: true)
+          run_clients_and_configure_setup(scope, create_snapshot: false, group => true)
```
akihikodaki, line 284: Why did you change this?

lib/engines/hcktest/hcktest.rb
```diff
@@ -295,8 +302,8 @@ def group_tests_by_config
       grouped_tests
     end
 
-    def run
-      upload_driver_package unless @driver_path.nil?
+    def run # rubocop:disable Metrics/AbcSize
```
akihikodaki, line 305: Refactor rather than disabling RuboCop.
In this case, I think you can remove the message "AutoHCK started in dump only mode"; it is obvious and we are not writing such a message for the auto run anyway.

lib/session.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/AbcSize,Lint/MissingCopEnableDirective
```
akihikodaki, line 3: Why?

lib/session.rb
```diff
@@ -0,0 +1,40 @@
...
+    def self.session_path(cli)
+      cli.test.latest_session ? "#{Config.read['workspace_path']}/latest" : cli.test.session
+    end
+
+    private_class_method def self.compose_session_json(options, workspace_path)
+      session = AutoHCK::Models::Session.new(
+        test: options.test,
+        common: options.common
+      )
+      session.common.workspace_path = workspace_path
```
akihikodaki, line 36: This has a side effect of modifying `options`, which is obscure and should be avoided.

lib/cli.rb
```diff
@@ -20,8 +21,15 @@ def initialize
...
     # class CommonOptions
-    class CommonOptions
-      attr_accessor :verbose, :config, :client_world_net, :id, :share_on_host_path, :workspace_path
+    class CommonOptions < T::Struct
+      include T::Sig
+
+      prop :verbose, T::Boolean, default: false
+      prop :config, T.nilable(String)
+      prop :client_world_net, T::Boolean, default: false
+      prop :id, T.nilable(Integer), default: 2
```
akihikodaki, line 30:
```suggestion
      prop :id, Integer, default: 2
```

akihikodaki (changes requested):

akihikodaki (changes requested):
The current implementation will leave an immature session if an error happens before clients get configured. Save the session after it is fully initialized (i.e., all clients are configured).

Jedoku:
> The current implementation will leave an immature session if an error happens before clients get configured. Save the session after it is fully initialized (i.e., all clients are configured).

But this is one of features, you can bring up VM`s for checking what's goin on and where it's crash

lib/cli.rb
```diff
@@ -77,15 +79,30 @@ def define_options(parser)
...
+      prop :gthb_context_prefix, T.nilable(String)
+      prop :gthb_context_suffix, T.nilable(String)
+      prop :playlist, T.nilable(String)
+      prop :select_test_names, T.nilable(String)
+      prop :reject_test_names, T.nilable(String)
+      prop :reject_report_sections, T::Array[String], default: []
+      prop :boot_device, T.nilable(String)
+      prop :allow_test_duplication, T.nilable(T::Boolean)
+      prop :manual, T.nilable(T::Boolean)
+      prop :package_with_playlist, T.nilable(T::Boolean)
```
kostyanf14, line 103: Don't see any reason to make boolean fields nilable. String or Array can be nilable because we have logic to check is field specified or not. But in the case of a boolean, we always check true or false.

akihikodaki:
> > The current implementation will leave an immature session if an error happens before clients get configured. Save the session after it is fully initialized (i.e., all clients are configured).
> 
> But this is one of features, you can bring up VM`s for checking what's goin on and where it's crash

The current code will make tests fail in an unpredictable manner instead of reproducing the past failure. It skips creating snapshots and configuration when restored so if some failure has happened in these operations, it will not reproduce the failure, but it will run the following tests with an immature configuration and pose different symptoms that are not useful to debug.

Jedoku:
It runs saved snapshots, and our VM's will be configured exactly as it was,
also on we can open studio and see what happens in previous session

akihikodaki:
> It runs saved snapshots, and our VM's will be configured exactly as it was, also on we can open studio and see what happens in previous session

At least the session should be saved at least after snapshots are created, and if the VMs died before configuration, we should configure them before continuing to run tests. 

The studio should have seen almost no activity at the early stage of execution before running tests; if it experienced a failure at the moment, it means the VM is so broken that it cannot boot or stay running for minutes. Such a scenario should be rare.

The easiest way to ensure the VMs are configured is simply to save the session after the configuration. It is sufficient to reproduce failures in tests.

lib/session.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module AutoHCK
+  class Session
```
akihikodaki, line 4: Make this module instead of class as instantiating it makes no sense.

lib/models/session.rb
```diff
@@ -0,0 +1,14 @@
...
+  module Models
+    class Session < T::Struct
+      extend T::Sig
+      extend JsonHelper
+
+      const :test, AutoHCK::CLI::TestOptions
+      const :common, AutoHCK::CLI::CommonOptions
+    end
+  end
+end
```
akihikodaki, line 14: What about letting `AutoHCK::CLI` inherit `T::Struct` so that we can remove this class?

lib/project.rb
```diff
@@ -161,15 +173,28 @@ def init_workspace
...
+    def add_latest_symlink
       begin
         File.delete("#{@config['workspace_path']}/latest")
       rescue Errno::ENOENT
         # firts run, no symlink to delete
       end
 
       File.symlink(@workspace_path, "#{@config['workspace_path']}/latest")
-      @setup_manager_type&.enter @workspace_path
     end
```
akihikodaki, line 198: I suggest merging this into `create_workspace`. Creating a workspace automatically makes it the latest so it makes sense to automatically update the symlink. The code in `init_workspace` will be more natural as you can write it to say "if being restored, restore the workspace; otherwise create a new one".

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -689,7 +690,7 @@ def run(scope, run_opts = nil)
         dump_commands
       else
         unless @configured
-          run_config_commands
+          run_config_commands unless @run_opts[:restored]
```
akihikodaki, line 693: Let's name the new option `configured` instead.
If it is named `restored`, this statement will mean: "if restored, the machine is already configured"; we already have such a logic in `HCKTest` so let the class contain the concern.
Theoretically, we can also instantiate `QemuMachine` without "restoring" in one session; whether restoration is happening or not does not matter for `QemuMachine`.

lib/project.rb
```diff
@@ -146,11 +146,23 @@ def check_run_termination
...
 
     def init_workspace
+      @restored = @options.test.session
```
akihikodaki, line 149: Avoid creating a new state by introducing a new instance variable if possible; you can define a `restored?` method instead.

akihikodaki (changes requested):

lib/cli.rb
```diff
@@ -2,240 +2,275 @@
...
+
+      @parser = @common.create_parser(@sub_parser)
     end
 
     def parse(args)
       left = @parser.order(args)
       @mode = left.shift
       @sub_parser[@mode]&.order!(left) unless @mode.nil?
+
+      restore_session if @test.latest_session || @test.session
```
akihikodaki, line 269: Instead of passing self to `AutoHCK::Session.load` and letting it modify, let's pass the path to the session.
Such a method may be called from `bin/auto_hck` and return an instance of loaded `CLI`, which will replace the whole existing `CLI` instance.

So instead of having these lines in `session.rb`:
```rb
      cli.test = session.test
      cli.common = session.common
```

...add a line like the following in `auto_hck`:
```rb
cli = Session.load(session_path)
```

You don't have to list each attribute of `CLI` this way.

akihikodaki (changes requested):

lib/cli.rb
```diff
@@ -2,234 +2,268 @@
...
-      end
-      # rubocop:enable Metrics/MethodLength
+  # class CLI
+  class CLI < T::Struct
+    extend AutoHCK::Models::JsonHelper
+    extend T::Sig
+
+    prop :test, T.nilable(TestOptions), default: nil
+    prop :common, T.nilable(CommonOptions), default: nil
+    prop :install, T.nilable(InstallOptions), default: nil
```
akihikodaki, line 248: Use `factory:` keyword argument documented in: https://sorbet.org/docs/tstruct

bin/auto_hck
```diff
@@ -9,6 +9,13 @@ module AutoHCK
   run do
     cli = CLI.new
     cli.parse(ARGV)
+    session_path = if cli.test.latest_session
+                     "#{Config.read['workspace_path']}/latest"
+                   elsif cli.test.session
```
akihikodaki, line 14: Simpler:

```suggestion
                   else
```

bin/auto_hck
```diff
@@ -9,6 +9,13 @@ module AutoHCK
...
+                     cli.test.session
+                   end
+
+    cli = Session.load(session_path, cli) if session_path
```
akihikodaki, line 18: Pass `cli.common.workspace_path` instead of `cli` to contain all code dealing with the original `cli` here (including the code deriving `session_path`).

lib/session.rb
```diff
@@ -0,0 +1,22 @@
...
+    def self.load(session_path, cli)
+      session = AutoHCK::CLI.from_json_file("#{session_path}/session.json")
+
+      session.common.workspace_path = cli.common.workspace_path
+      session.test.session = session_path
+      session.test.manual = true
+      session
+    end
+  end
+end
```
akihikodaki, line 22: Better to merge this into `AutoHCK::CLI` as its methods all deal with that class.

akihikodaki (changes requested):

bin/auto_hck
```diff
@@ -10,6 +10,11 @@ module AutoHCK
     cli = CLI.new
     cli.parse(ARGV)
 
+    if cli.test.latest_session || cli.test.session
+      session_path = cli.test.latest_session ? "#{Config.read['workspace_path']}/latest" : cli.test.session
+      cli = Session.load(session_path, cli)
+    end
```
akihikodaki, line 16: Simpler:

```suggestion
    session_path = cli.test.latest_session ? "#{Config.read['workspace_path']}/latest" : cli.test.session
    cli = Session.load(session_path, cli) if session_path
```

akihikodaki (changes requested):

akihikodaki (changes requested):
See the unresolved comments.

lib/engines/hcktest/hcktest.rb
akihikodaki: My question is not about session restoration, but about the normal case that does not involve session restoration. This change will affect all scenarios since it does not check `project.restored?`; is that OK?

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -688,9 +688,9 @@ def run(scope, run_opts = nil)
       if @run_opts[:dump_only]
         dump_commands
       else
-        unless @configured
+        unless run_opts[:configured]
```
akihikodaki, line 691:
```suggestion
        unless @run_opts[:configured]
```

lib/cli.rb
```diff
@@ -2,234 +2,264 @@
...
+    extend AutoHCK::Models::JsonHelper
+    extend T::Sig
+
+    prop :test, T.nilable(TestOptions), factory: -> { TestOptions.new }
+    prop :common, T.nilable(CommonOptions), factory: -> { CommonOptions.new }
+    prop :install, T.nilable(InstallOptions), factory: -> { InstallOptions.new }
+    prop :mode, T.nilable(String), default: nil
+
+    attr_reader :install, :mode
+    attr_accessor :common, :test
```
akihikodaki, line 252: I don't think they are necessary.

lib/session.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module AutoHCK
+  class Session < T::Struct
+    extend T::Sig
+    extend Models::JsonHelper
+
+    prop :common, CommonOptions
+    prop :test, TestOptions
+    prop :install, InstallOptions
+    prop :mode, T.nilable(String), default: nil
```
akihikodaki, line 12: At https://github.com/HCK-CI/AutoHCK/pull/487#discussion_r1981046821, I suggested adding a property for `CLI` instead of properties that `CLI` has.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -688,9 +688,9 @@ def run(scope, run_opts = nil)
       if @run_opts[:dump_only]
         dump_commands
       else
-        unless @configured
+        unless @run_opts[:configured]
           run_config_commands
-          @configured = true
+          @run_opts[:configured] = true
```
akihikodaki, line 693: The changes in `qemu_machine.rb` break the scenario where the `run` method is called multiple times without `configured` specified.

lib/cli.rb
```diff
@@ -2,234 +2,263 @@
...
+  class CLI < T::Struct
+    extend AutoHCK::Models::JsonHelper
+    extend T::Sig
+
+    prop :test, T.nilable(TestOptions), factory: -> { TestOptions.new }
+    prop :common, T.nilable(CommonOptions), factory: -> { CommonOptions.new }
+    prop :install, T.nilable(InstallOptions), factory: -> { InstallOptions.new }
+    prop :mode, T.nilable(String), default: nil
+
+    attr_reader :common, :install, :test, :mode
```
akihikodaki, line 251: I don't think these are necessary.

lib/session.rb
```diff
@@ -0,0 +1,27 @@
...
+  class Session < T::Struct
+    extend T::Sig
+    extend Models::JsonHelper
+
+    prop :cli, CLI
+
+    sig { params(workspace_path: String, cli: CLI, logger: AutoHCK::MultiLogger).void }
+    def self.save_cli(workspace_path, cli, logger)
+      File.write("#{workspace_path}/session.json", cli.serialize.to_json)
+      logger.info("Session saved to #{workspace_path}/session.json")
```
akihikodaki, line 14: It's not saving `Session` but saving `CLI`.

akihikodaki (changes requested):

lib/session.rb
Jedoku: Indeed

kostyanf14:
```
    def self.save(workspace_path, logger)
      File.write("#{workspace_path}/session.json", self.to_json)
```

lib/session.rb
```diff
@@ -0,0 +1,34 @@
...
+      File.write("#{workspace_path}/session.json", serialize.to_json)
+      logger.info("Session saved to #{workspace_path}/session.json")
+    end
+
+    sig { params(session_path: String, workspace_path: T.nilable(String)).returns(Session) }
+    def load(session_path, workspace_path)
+      session = Session.from_json_file("#{session_path}/session.json")
+      @cli = load_cli(session, session_path, workspace_path)
+      session
+    end
```
akihikodaki, line 22: Make this a class method as it creates a new instance.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -684,6 +684,7 @@ def run(scope, run_opts = nil)
 
       process_device_commands
       normalize_lists
+      @configured ||= @run_opts[:configured]
```
akihikodaki, line 687: Do not ignore `@run_opts[:configured]` when the caller set it to `false`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -684,6 +684,7 @@ def run(scope, run_opts = nil)
 
       process_device_commands
       normalize_lists
+      @configured = @run_opts[:configured] if [true, false].include?(@run_opts[:configured])
```
akihikodaki, line 687: To make this simpler and torelate other truthy and falsy values:

```suggestion
      @configured = @run_opts[:configured] if @run_opts.key?(:configured)
```

lib/engines/hcktest/tests.rb
```diff
@@ -34,7 +34,6 @@ def initialize(client, support, project, target, tools)
 
     def update_tests(log: false)
       retries ||= 0
-
```
akihikodaki, line 37: This changr is irrelevant.

akihikodaki (changes requested):

lib/project.rb
```diff
@@ -28,6 +30,10 @@ def initialize(scope, options)
       scope << self
     end
 
+    def initialize_json_override
+      Json.update_json_override(@options.common.config) unless @options.common.config.nil?
+    end
```
akihikodaki, line 35: This method is too trivial. I would rather fix the Rubocop warning with `initialize` by removing `@id`.

akihikodaki (changes requested):

akihikodaki (changes requested):
Please look at unresolved comments:
- https://github.com/HCK-CI/AutoHCK/pull/487#discussion_r2015972343
- https://github.com/HCK-CI/AutoHCK/pull/487#discussion_r2026245429
- https://github.com/HCK-CI/AutoHCK/pull/487#discussion_r2036861966

lib/engines/hcktest/hcktest.rb
```diff
@@ -184,13 +184,14 @@ def configure_setup_and_synchronize
...
       @clients.each_value(&:keep_snapshot)
+      @project.save_session unless @project.restored?
     end
 
     def run_clients_and_configure_setup(scope, **opts)
       retries = 0
       begin
         scope.transaction do |tmp_scope|
-          run_clients tmp_scope, keep_alive: true, **opts
+          run_clients tmp_scope, keep_alive: true, **opts, **init_vm_options
```
akihikodaki, line 194: `**init_vm_options` is already applied by `run_tests_without_config`.

lib/engines/hcktest/hcktest.rb
```diff
@@ -240,7 +241,6 @@ def pause_run
 
     def prepare_tests
       client, support = @clients.values
-
```
akihikodaki, line 243: This change is irrelevant.

lib/session.rb
```diff
@@ -0,0 +1,31 @@
...
+    sig { params(session_path: String, workspace_path: T.nilable(String)).returns(Session) }
+    def self.load(session_path, workspace_path)
+      session = Session.from_json_file("#{session_path}/session.json")
+      load_cli(session, session_path, workspace_path)
+      session
+    end
+
+    sig { params(session: Session, session_path: String, workspace_path: T.nilable(String)).void }
+    def self.load_cli(session, session_path, workspace_path)
+      T.must(session.cli.common).workspace_path = workspace_path
```
akihikodaki, line 26: `T.must` is no longer necessary.

akihikodaki (changes requested):

akihikodaki (changes requested):

akihikodaki (changes requested):
Two of the last comments are not addressed.

lib/engines/hcktest/hcktest.rb
```diff
@@ -249,12 +250,24 @@ def prepare_tests
         @project.logger.info('HLK test target is not defined, allow in manual mode')
       end
 
+      @tests.verify_target if @project.restored?
```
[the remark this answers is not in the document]
Jedoku: I actually don't clearly remember it but, i think it was some error relatet to target creation back then
Anyway it's works without it now

lib/project.rb
```diff
@@ -181,7 +186,22 @@ def init_workspace
       end
 
       File.symlink(@workspace_path, "#{@config['workspace_path']}/latest")
-      @setup_manager_type&.enter @workspace_path
+    end
+
+    def restore_workspace
+      @workspace_path = @options.test.session
+
+      raise AutoHCKError, 'Workspace path does not exist could not load session' unless File.directory?(@workspace_path)
```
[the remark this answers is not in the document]
akihikodaki: Rescue an error only when restored. Checking `File.directory?(session_path)` results in a duplicate file operation and a "time-of-check to time-of-use" race condition, which is why I suggest rescuing one.

lib/engines/hcktest/hcktest.rb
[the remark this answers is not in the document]
akihikodaki: The new version removed the condition, but `@tests.verify_target` is still there.

akihikodaki (dismissed):
I replied to the comments so please check them out.

lib/project.rb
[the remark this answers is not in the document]
Jedoku: It actually will be rised by OpenJsonError if it's does't exist, I dont want to rescue it's doesn't make sense for me we can have a little double operation or you can sugget another one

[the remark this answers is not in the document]
Jedoku: Look, we cant catch Errno::ENOENT error here Cause its already catched by Json helper and our rescue will be ignored
So we can remove this rising of error and will see that session.JSON does not exist ant it will be unclear that workspace_path doesn't exist

lib/engines/hcktest/hcktest.rb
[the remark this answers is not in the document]
Jedoku:
```
I, [2025-05-29T09:11:23.184397 #1941716]  INFO -- CmdRun PID:1943108: Command finished with code 0
F, [2025-05-29T09:11:23.184889 #1941716] FATAL -- : (NoMethodError) undefined method `[]' for nil
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/tests.rb:61:in `update_tests'
   -- /home/jedoku/HCK/repos/AutoHCK.git/vendor/bundle/ruby/3.3.0/gems/sorbet-runtime-0.5.12020/lib/types/private/methods/call_validation.rb:282:in `bind_call'
   -- /home/jedoku/HCK/repos/AutoHCK.git/vendor/bundle/ruby/3.3.0/gems/sorbet-runtime-0.5.12020/lib/types/private/methods/call_validation.rb:282:in `validate_call'
   -- /home/jedoku/HCK/repos/AutoHCK.git/vendor/bundle/ruby/3.3.0/gems/sorbet-runtime-0.5.12020/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:253:in `prepare_tests'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:281:in `block in run_tests_without_config'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:29:in `block (2 levels) in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `block in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:279:in `run_tests_without_config'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:271:in `block in auto_run'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:29:in `block (2 levels) in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `block in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:267:in `auto_run'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/engines/hcktest/hcktest.rb:329:in `run'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/project.rb:48:in `run'
   -- bin/auto_hck:30:in `block (2 levels) in <module:AutoHCK>'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:29:in `block (2 levels) in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:28:in `block in open'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `handle_interrupt'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auxiliary/resource_scope.rb:27:in `open'
   -- bin/auto_hck:27:in `block in <module:AutoHCK>'
   -- /home/jedoku/HCK/repos/AutoHCK.git/lib/auto_hck.rb:14:in `run'
   -- bin/auto_hck:9:in `<module:AutoHCK>'
   -- bin/auto_hck:8:in `<main>'
D, [2025-05-29T09:11:23.185073 #1941716] DEBUG -- : Closing AutoHCK project
```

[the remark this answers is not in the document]
Jedoku: That's why it was there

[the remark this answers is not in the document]
Jedoku: `I, [2025-05-29T09:10:41.233651 #1941716]  INFO -- : HLK test target is not defined, allow in manual mode
I, [2025-05-29T09:10:41.234026 #1941716]  INFO -- : Exiting HLKClient CL1
I, [2025-05-29T09:10:41.234185 #1941716]  INFO -- : Sending powerdown signal to QemuMachine0002_CL01 via QMP
I, [2025-05-29T09:11:08.072050 #1941716]  INFO -- CmdRun PID:1942171: Command finished with code 0`

lib/project.rb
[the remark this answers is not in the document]
akihikodaki: By the way, do we really need to check if it was `Errno::ENOENT`? I guess you can always tell the user that loading a session failed, whatever the error was.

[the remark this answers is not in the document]
Jedoku: but its only related do json file, its unclear that whole workspace path does not exist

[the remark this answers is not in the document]
akihikodaki: I think the problem here is that `AutoHCK::OpenJsonError` is redundant. It will be clear if we stop raising that exception for `Errno::ENOENT`.

lib/engines/hcktest/hcktest.rb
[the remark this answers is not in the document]
akihikodaki: The log indicates the underlying problem is that `HCKClient#add_target_to_project` is not called because 
`configure_and_synchronize_clients` is skipped. Letting `configure_and_synchronize_clients` call `HCKClient#add_target_to_project` would be a proper fix.

lib/project.rb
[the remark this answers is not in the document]
Jedoku: >  If the whole workspace path doesn't exist, the user specified a wrong path. If only the JSON file is missing, the user specified a path to a wrong directory. Either way, the user needs to check if they typed the correct path.

That's how it works right now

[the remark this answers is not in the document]
akihikodaki: > > If the whole workspace path doesn't exist, the user specified a wrong path. If only the JSON file is missing, the user specified a path to a wrong directory. Either way, the user needs to check if they typed the correct path.
> 
> That's how it works right now

Yes, but `File.directory?(session_path)` is unnecessary.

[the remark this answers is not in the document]
Jedoku: done

lib/engines/hcktest/hcktest.rb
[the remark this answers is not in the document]
akihikodaki: Right. We need another method to call `search_target` instead.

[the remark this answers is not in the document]
Jedoku: What's wrong with verify_target?

kostyanf14:
Replaced https://github.com/HCK-CI/AutoHCK/pull/977

PR #493, opened 2024-10-09
Viomem fix

kostyanf14:


YanVugenfirer (approved):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -176,6 +176,13 @@ def initialize(options)
       @devices_list.flatten!
       @devices_list.compact!
       @device_infos = load_devices
+      process_device_options
+    end
+
+    def process_device_options
+      @device_infos.each do |device_info|
+        @pluggable_memory_gb += device_info.pluggable_memory_gb
+      end
```
akihikodaki, line 185: I prefer this to be merged into `memory_replacement_map` to remove the instance variable.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
kostyanf14: We discussed previously, that `pluggable_memory_gb` should be a separate value and autohck should calculate max memory size

YanVugenfirer (approved):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -310,10 +309,15 @@ def machine_replacement_map
     def memory_replacement_map
       memory_gb = option_config('memory_gb')
 
+      pluggable_memory_gb = 0
+      @device_infos.each do |device_info|
+        pluggable_memory_gb += device_info.pluggable_memory_gb
+      end
```
akihikodaki, line 315: Use `Array#sum`.

akihikodaki (changes requested):

akihikodaki (approved):

PR #494, opened 2024-10-10
Generalize DiffChecker class and tool

kostyanf14:
Add extra cli option to allow to use '*' during triggers check.


akihikodaki:
I want to know the use case for this change.

akihikodaki (approved):

YanVugenfirer (approved):

PR #499, opened 2024-10-21
SVVP: Fix starting SVVP in manual mode without tests

kostyanf14:


lib/engines/hcktest/hcktest.rb
```diff
@@ -239,7 +239,6 @@ def prepare_tests
         raise EngineError, 'HLK test target is not defined' unless @project.options.test.manual
 
         @project.logger.info('HLK test target is not defined, skipping tests loading in manual mode')
```
akihikodaki, line 241: This message sounds misleading.

akihikodaki (changes requested):

akihikodaki (approved):

PR #506, opened 2024-10-24
Skip selected as a boot device driver

Jedoku:


lib/engines/hcktest/hcktest.rb
```diff
@@ -86,15 +86,22 @@ def read_driver(driver)
       Models::Driver.from_json_file(driver_json, @logger)
     end
 
+    def driver_installed?(driver)
+      return false unless @project.options.test.svvp
```
akihikodaki, line 90: What happens if `@project.options.test.svvp && driver.device == @project.options.test.boot_device`?

akihikodaki (changes requested):

kostyanf14 (commented):
@akihikodaki 

Currently, we have broken QEMU CLI for SVVP, not sure this is a bug or not, because we add the same device 2 times. The old problem was during driver installation, pnputil fail and auto_hck fail

akihikodaki:
> Currently, we have broken QEMU CLI for SVVP, not sure this is a bug or not, because we add the same device 2 times. The old problem was during driver installation, pnputil fail and auto_hck fail

In that case, let's remove the statement `return false unless @project.options.test.svvp`. It does not seem to be the correct way to determine whether the driver is already installed for SVVP. We need to figure out that later.

lib/engines/hcktest/hcktest.rb
```diff
@@ -86,10 +86,15 @@ def read_driver(driver)
       Models::Driver.from_json_file(driver_json, @logger)
     end
 
+    def driver_installed?(driver)
+      driver.device == @project.options.test.boot_device
+    end
```
akihikodaki, line 91: Let's merge this method into `find_drivers`; this method is now just one statement.

akihikodaki (changes requested):

akihikodaki (approved):

kostyanf14 (approved):

PR #516, opened 2024-11-06
QemuHCK: Fix missing iommu_device_param on device

kostyanf14:


YanVugenfirer (approved):

YanVugenfirer (approved):

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -353,6 +352,11 @@ def device_machine_options
       @device_infos.map(&:machine_options).flatten.compact
     end
 
+    sig { returns(T::Array[String]) }
+    def device_iommu_device_param
+      @device_infos.map(&:iommu_device_param).flatten.compact
```
akihikodaki, line 357: Use `flat_map`.

akihikodaki (changes requested):

lib/setupmanagers/qemuhck/qemu_machine.rb
akihikodaki: No, `iommu_device_param` is not an Array but just a mere String so we can just omit `flatten`.

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -353,6 +352,11 @@ def device_machine_options
       @device_infos.map(&:machine_options).flatten.compact
     end
 
+    sig { returns(T::Array[String]) }
+    def device_iommu_device_param
+      @device_infos.map(&:iommu_device_param).compact
```
akihikodaki, line 357: Use `filter_map` to make this even shorter.

akihikodaki (changes requested):

akihikodaki (approved):

PR #517, opened 2024-11-07
Misc kits logic update

kostyanf14:


akihikodaki (changes requested):

spec/kit_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        # a size that differs from the data chunk size.
+        # According to docs a Curl::Err::AbortedByCallbackError should be raised
+        # Really, Curl::Err::WriteError is raised.
+        writed_size < 32_769 ? data.size : 0
+      end
+
+      begin
+        easy.perform
+      rescue Curl::Err::WriteError
+      end
```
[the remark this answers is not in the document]
kostyanf14: >I'm not enthusiastic about having another dependency on curb

We already have it.

Not sure that we got more simpler code with `net-http/Faraday`.  First of all `net-http/Faraday` does not perform HTTP redirection automatically and we should perform it manually or add new gem  `faraday_middleware`.
```
3.2.2 :019 > resp = Faraday.get('https://go.microsoft.com/fwlink/?linkid=2250318')
 => 
#<Faraday::Response:0x00007ef0036c7470
... 
3.2.2 :020 > resp.status
 => 302 
3.2.2 :021 > resp.body
 => "" 
 ```
 
You talk about `Net::HTTPHeader#range` but server can ignore it and we got the full file. This is not our servers and we don't know how Microsoft process this as a real file URL are often temporary. 
Even if server support  `Net::HTTPHeader#range` for which request we should specify it?

File: lib/engines/hckinstall/hckinstall.rb [hunk 2]
```diff
@@ -93,13 +93,11 @@ def init_class_variables
     end
 
     def prepare_extra_sw
-      [@kit_info, @project.engine_platform].each do |source|
-        next if source['extra_software'].nil?
+      extra_software = [@kit_info.extra_software, @project.engine_platform['extra_software']].flatten.compact
```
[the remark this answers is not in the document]
kostyanf14: Fixed. `@project.engine_platform['extra_software']` is array to.

spec/kit_spec.rb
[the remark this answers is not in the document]
akihikodaki: httpclient worked nice so let's use it:

```Ruby
HTTPClient.get_content('https://go.microsoft.com/fwlink/?linkid=2289283') { p _1.size; throw 'foo' }
```

`throw` is used here to abort downloading.

spec/kit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      writed_size = 0
+      begin
+        HTTPClient.get_content(kit_url) do |data_chunk|
+          temp_file << data_chunk
+          writed_size += data_chunk.size
+          raise HTTPClientDownloadNeededPart if writed_size > 32_769
+        end
+      rescue HTTPClientDownloadNeededPart
+      end
```
akihikodaki, line 26: Use `throw` and `catch`. They are mechanisms to escape the execution locally.

spec/kit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+describe 'kit_spec' do
+  Dir['./lib/engines/hckinstall/kits/*.json'].each do |json_file|
+    kit_url = AutoHCK::Models::Kit.from_json_file(json_file).download_url
+    if kit_url.nil?
+      print("No download URL for #{json_file}\n")
+      next
+    end
+
+    it "Kit for #{json_file} can be downloaded" do
+      temp_file = StringIO.new
```
akihikodaki, line 16: - Use a mutable `String`, which `StringIO` wraps. We don't need a `IO` interface here.
- It is not a file. In general I recommend using an existing terminology. For example `content` as it's used for `HTTPClient.get_content`.

spec/kit_spec.rb
```diff
@@ -0,0 +1,38 @@
+require_relative '../lib/all'
+require 'stringio'
+require 'httpclient'
```
akihikodaki, line 3: `../lib/all` automatically requires it.

spec/kit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class HTTPClientDownloadNeededPart < StandardError; end
+
+describe 'kit_spec' do
+  Dir['./lib/engines/hckinstall/kits/*.json'].each do |json_file|
+    kit_url = AutoHCK::Models::Kit.from_json_file(json_file).download_url
+    if kit_url.nil?
+      print("No download URL for #{json_file}\n")
```
akihikodaki, line 11: I'm worried that this may get printed for all test runs even when you run irrelevant tests.

spec/kit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      next
+    end
+
+    it "Kit for #{json_file} can be downloaded" do
+      temp_file = StringIO.new
+
+      writed_size = 0
```
akihikodaki, line 18: This variable is not necessary as we can use `temp_file.size`.

akihikodaki (changes requested):

spec/kit_spec.rb
kostyanf14: why?
is this better to use it + pending in this case?

spec/kit_spec.rb
```diff
@@ -0,0 +1,30 @@
+require_relative '../lib/all'
+
+class HTTPClientDownloadNeededPart < StandardError; end
```
akihikodaki, line 3: This is not necessary; check https://docs.ruby-lang.org/en/3.3/Kernel.html#method-i-catch

spec/kit_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    if kit_url.nil?
+      print("No download URL for #{json_file}\n")
+      next
+    end
+
+    it "Kit for #{json_file} can be downloaded" do
+      content = String.new
+
+      catch HTTPClientDownloadNeededPart do
+        HTTPClient.get_content(kit_url) do |data_chunk|
```
akihikodaki, line 17: I suggest renaming `data_chunk` to `chunk`; the word "data" tells nothing.

akihikodaki (changes requested):

spec/kit_spec.rb
```diff
@@ -0,0 +1,27 @@
+require_relative '../lib/all'
+
+describe 'kit_spec' do
+  Dir['./lib/engines/hckinstall/kits/*.json'].each do |json_file|
+    kit_url = AutoHCK::Models::Kit.from_json_file(json_file).download_url
+    if kit_url.nil?
+      next
+    end
```
akihikodaki, line 8:
```suggestion
    next if kit_url.nil?
```

spec/kit_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it "Kit for #{json_file} can be downloaded" do
+      content = String.new
+
+      catch 1 do
+        HTTPClient.get_content(kit_url) do |chunk|
+          content << chunk
+          throw 1 if content.size > 32_773
```
akihikodaki, line 16: Let `catch` create a tag for us instead of using 1. Using 1 is unsafe as httpclient may, in theory, internally have `catch 1`.

akihikodaki (changes requested):

spec/kit_spec.rb
akihikodaki: I found we can just use `break` here... my bad assuming `break` doesn't work.

akihikodaki (changes requested):
See: https://github.com/HCK-CI/AutoHCK/pull/517#discussion_r1841825927

akihikodaki (approved):

PR #568, opened 2025-01-30
RHELMISC-7213: Dropbox uploader create folder but does not create share folder

Jedoku:


lib/resultuploaders/dropbox/dropbox.rb
```diff
@@ -118,13 +118,11 @@ def connect
...
-        @dropbox.create_folder(@path)
-        @dropbox.share_folder(@path)
-        @url = "#{@dropbox.create_shared_link_with_settings(@path).url}&lst="
-        @logger.info("Dropbox project folder created: #{@url}")
-      end
+      @path = "/#{@repo}/CI/#{@tag}-#{@timestamp}"
+      handle_exceptions(__method__) { @dropbox.create_folder(@path) }
+      handle_exceptions(__method__) { @dropbox.share_folder(@path) }
+      @url = "#{@dropbox.create_shared_link_with_settings(@path).url}&lst="
+      @logger.info("Dropbox project folder created: #{@url}")
```
kostyanf14, line 125: `dropbox.create_shared_link_with_settings` is also a Dropbox API that can fail, so should be in `handle_exceptions` too.

lib/resultuploaders/dropbox/dropbox.rb
```diff
@@ -118,13 +118,11 @@ def connect
...
-      handle_exceptions(__method__) do
-        @path = "/#{@repo}/CI/#{@tag}-#{@timestamp}"
-        @dropbox.create_folder(@path)
-        @dropbox.share_folder(@path)
-        @url = "#{@dropbox.create_shared_link_with_settings(@path).url}&lst="
-        @logger.info("Dropbox project folder created: #{@url}")
-      end
+      @path = "/#{@repo}/CI/#{@tag}-#{@timestamp}"
+      handle_exceptions(__method__) { @dropbox.create_folder(@path) }
+      handle_exceptions(__method__) { @dropbox.share_folder(@path) }
```
kostyanf14, line 123: When create_folder or share_folder fails we get absolutely the same error and this will be hard to investigate. Split into different functions to have different `__method__` in logs.

kostyanf14 (changes requested):

akihikodaki (changes requested):
Please explain what this change solves and how in the commit message.

kostyanf14 (approved):

akihikodaki (changes requested):
This change will try to share, create, and print a link to the folder even if the creation of the folder fails.

akihikodaki (changes requested):
It still tries to share and create a link to the folder even if the creation of the folder fails.

kostyanf14:
> It still tries to share and create a link to the folder even if the creation of the folder fails.

@akihikodaki I don't think this is a problem. Sharing will fail in this case too. This is more complicated to make sure that folder is created before trying to share it. 

What are your suggestions?

akihikodaki:
> > It still tries to share and create a link to the folder even if the creation of the folder fails.
> 
> @akihikodaki I don't think this is a problem. Sharing will fail in this case too. This is more complicated to make sure that folder is created before trying to share it.
> 
> What are your suggestions?

Add a variable to tell if a folder is already created or not instead of having two `handle_exceptions`. Sharing will fail indeed, but it's better to avoid trying that.

akihikodaki (changes requested):
In general, rescue an error instead of detecting an error pattern before execution. In this case, rescue `DropboxApi::Errors::FolderConflictError`.
It saves I/O and avoids a race condition.

akihikodaki (approved):

kostyanf14 (approved):

PR #617, opened 2025-03-10
RHELMISC-9551: Update cleanup script according latest changes in AutoHCK

Jedoku:


bin/cleanup_workspace
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+Dir.chdir(File.dirname(__dir__))
+
+require_relative '../lib/auto_hck'
+
+module AutoHCK
+  MONTH_IN_SECS = 2_592_000
```
akihikodaki, line 9: Move this into `CleanupWorkspace` to avoid name collision.

bin/cleanup_workspace
```diff
@@ -0,0 +1,68 @@
...
+      pattern = File.join(
+        workspace_path,
+        '**',
+        '[0-9][0-9][0-9][0-9]_' \
+        '[0-9][0-9]_' \
+        '[0-9][0-9]_' \
+        '[0-9][0-9]_' \
+        '[0-9][0-9]_' \
+        '[0-9][0-9]'
+      )
```
akihikodaki, line 34: This pattern is too broad and dangerous.

bin/cleanup_workspace
```diff
@@ -0,0 +1,68 @@
...
+      config = AutoHCK::Config.read
+      workspace_path = config['workspace_path']
+      workspace_folders(workspace_path).each do |test_run|
+        delete(test_run) if delete?(test_run.split('/').last)
+      end
+    end
+  end
+
+  run do
+    extend CleanupWorkspace
```
akihikodaki, line 49: This pollutes `Object`.

akihikodaki (changes requested):

bin/cleanup_workspace
```diff
@@ -0,0 +1,77 @@
...
+
+module AutoHCK
+  class WorkspaceCleaner
+    def initialize(cleanup_time)
+      @cleanup_time = cleanup_time || 2_592_000 # Default to 30 days
+    end
+
+    def self.cleanup(cleanup_time)
+      new(cleanup_time).cleanup
+    end
```
akihikodaki, line 16: This method is too trivial

bin/cleanup_workspace
```diff
@@ -0,0 +1,77 @@
...
+      puts "Deleting #{test_run}"
+      FileUtils.remove_entry_secure(test_run, true)
+    rescue StandardError => e
+      puts "Error deleting #{test_run}: #{e.message}"
+    end
+
+    def workspace_folders(workspace_path)
+      timestamp_regex = /^\d{4}_\d{2}_\d{2}_\d{2}_\d{2}_\d{2}$/
+
+      Dir.glob(File.join(workspace_path, '**', '*')).select do |path|
```
akihikodaki, line 33: Don't use `**` to avoid recursing into each workspace.

bin/cleanup_workspace
```diff
@@ -0,0 +1,77 @@
...
+    def cleanup
+      config = AutoHCK::Config.read
+      workspace_path = config['workspace_path']
+      folders = workspace_folders(workspace_path)
+
+      puts "Found #{folders.count} timestamp folders in workspace"
+
+      deleted_count = 0
+      folders.each do |test_run|
+        if delete?(File.basename(test_run))
```
akihikodaki, line 47: Simpler:

```suggestion
        next unless delete?(File.basename(test_run))
```

akihikodaki (changes requested):

bin/cleanup_workspace
Jedoku: I make it this way for better reading of this `WorkspaceCleaner.cleanup(cleanup_time)`

akihikodaki: I don't think it's significant enough. If this method is not present, you will need write: `WorkspaceCleaner.new.cleanup(cleanup_time)`. I don't think the presence of `.new` matters.

bin/cleanup_workspace
```diff
@@ -0,0 +1,73 @@
...
+
+module AutoHCK
+  class WorkspaceCleaner
+    def initialize(cleanup_time)
+      @cleanup_time = cleanup_time || 2_592_000 # Default to 30 days
+    end
+
+    def delete?(timestamp)
+      test_time = Time.strptime(timestamp, '%Y_%m_%d_%H_%M_%S')
+      Time.now - test_time > @cleanup_time.to_i
```
akihikodaki, line 16: Why `to_i` is needed here?

bin/cleanup_workspace
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def delete(test_run)
+      puts "Deleting #{test_run}"
+      FileUtils.remove_entry_secure(test_run, true)
+    rescue StandardError => e
+      puts "Error deleting #{test_run}: #{e.message}"
```
akihikodaki, line 23:
```suggestion
      puts "Error deleting #{test_run}: #{e}"
```

...should have the same effect with less code.

bin/cleanup_workspace
```diff
@@ -0,0 +1,73 @@
...
+
+    def cleanup
+      config = AutoHCK::Config.read
+      workspace_path = config['workspace_path']
+      folders = workspace_folders(workspace_path)
+
+      puts "Found #{folders.count} timestamp folders in workspace"
+
+      deleted_count = 0
+      folders.each do |test_run|
```
akihikodaki, line 42: I suggest using `Array#count`.

akihikodaki (changes requested):

bin/cleanup_workspace
```diff
@@ -0,0 +1,73 @@
...
+
+    def cleanup
+      config = AutoHCK::Config.read
+      workspace_path = config['workspace_path']
+      folders = workspace_folders(workspace_path)
+
+      puts "Found #{folders.count} timestamp folders in workspace"
+
+      deleted_count = 0
+      folders.count do |test_run|
```
akihikodaki, line 42: Use `Array#count` to get the deleted count.

akihikodaki (changes requested):

bin/cleanup_workspace
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  run do
+    cleanup_time = nil
+
+    parser = OptionParser.new do |opts|
+      opts.banner = 'Usage: bin/cleanup_workspace [options]'
+
+      opts.on('-t', '--time <time in seconds', Integer,
```
akihikodaki, line 56: I guess the closing `>` is missing here:

```suggestion
      opts.on('-t', '--time <time in seconds>', Integer,
```

akihikodaki (changes requested):

akihikodaki (approved):

PR #629, opened 2025-03-18
RFC: Fixes for run the same test several times

kostyanf14:


lib/engines/hcktest/tests.rb
```diff
@@ -226,8 +235,24 @@ def print_tests_stats
...
+      # so let's wait for result, but this is not error if real result
+      # will not be received
+      last_result = {}
+      (time_to_seconds(RUNNING_TEST_TIMEOUT) / 60).to_i.times do
+        last_result = last_test_result(test['id'])
+
+        break if last_result['status'] != 'Running'
+
+        sleep 60
+      end
```
akihikodaki, line 253: Why do we want to poll here while `handle_test_running` already polls?

lib/engines/hcktest/tests.rb
```diff
@@ -208,7 +212,12 @@ def test_finished?(test)
     end
 
     def done_tests
-      @tests.select { |test| test_finished?(test) }
+      # When test is running more than ones HLK reports test['status'] = PASS/FAIL
```
akihikodaki, line 215:
```suggestion
      # When test is running more than once HLK reports test['status'] = PASS/FAIL
```

I guess we need to ask ChatGPT to correct this comment.

lib/engines/hcktest/tests.rb
```diff
@@ -401,10 +427,23 @@ def apply_filters
     end
 
     def check_new_finished_tests
-      return unless new_done.any?
+      @logger.debug('Checking new finished tests')
+      new_done_tests = new_done
```
akihikodaki, line 431: Is `new_done` still necessary?

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
kostyanf14: yes. 

In case someone manually opens HLK studio and queue test. We do this on CI when we see some unexpected fails and CI is still running. In this case `@last_queued_id` will be `nil`.

akihikodaki: Won't the manually retried test stay "FAIL"? I think it's better to generalize the logic to check all tests that have new results instead of checking only the last queued test.

kostyanf14: this is why new_done is needed

scenario 1 normal run:
```
last done = [{ test=>'name1', result => 'fail'}]
done_tests  = [{ test=>'name1', result => 'fail'},  { test=>'name2', result => 'pass'}]
done_tests - last_done =  [{ test=>'name2', result => 'pass'}]
```

scenario 2 normal run + manual retry via VNC:
```
last done = [{ test=>'name1', result => 'fail'}]
done_tests  = [{ test=>'name1', result => 'pass'},  { test=>'name2', result => 'pass'}]
done_tests - last_done =   [{ test=>'name1', result => 'pass'},  { test=>'name2', result => 'pass'}]
```

scenario 3 test duplication:
```
last done = [{ test=>'name1', result => 'pass'}]
done_tests  = [{ test=>'name1', result => 'pass'}]
done_tests - last_done =   []
```


To cover scenario 3, I added `last_done_tests` logic
Also `last_done_tests` will duplicate scenario 1, that is the reason for `uniq` ((new_done_tests + last_done_tests).uniq)
To cover scenario 2 we definitely need `new_done` function because AutoHCK can't detect what a user do manually.

akihikodaki: What about removing the automatically retried test from `last_done` instead of adding the `last_done_tests` logic to cover scenario 3?

akihikodaki: Storing `last_test_results` sounds good to me.

kostyanf14: but we still have a question about how to get `done_tests_results`. We need to request tests_results for each test. This is a lot of API calls

akihikodaki: `last_test_result` already calls `list_test_results`  so it's fine I guess?

kostyanf14: it gets results for one test, but we need them for all tests.

We can write new toolsHLK command `list_all_results` but even HLK loads results only for a specific test when you open it.

akihikodaki: Writing a command may be an option. I don't see a reason that makes `list_all_results` prohibitively slow.

kostyanf14:
replaced with https://github.com/HCK-CI/AutoHCK/pull/631

PR #630, opened 2025-03-19
RHELMISC-9548: Differentiate GUI and CORE platforms on the ISO config

Jedoku:


lib/engines/hckinstall/iso.json
```diff
@@ -9,5 +9,16 @@
...
     }
+  }, 
+  "Win2022x64_GUI": {
+    "path": "en_windows_server_2022_x64_dvd_7f6f8e3e.iso",
+    "studio": {
+      "windows_image_names": "Windows Server 2022 SERVERSTANDARD",
+      "product_key": "BBBBB-BBBBB-BBBBB-BBBBB-BBBBB"
+    },
+    "client": {
+      "windows_image_names": "Windows Server 2022 SERVERSTANDARDCORE",
```
akihikodaki, line 20: It has `CORE` in the image name though the key is suffixed with `_GUI` instead of `_CORE`. Are you sure this is right?

Jedoku: It's RFC version, i'll fix it, also we need to deciede what to do with our platform json's

lib/engines/hckinstall/iso.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Win2019x64": {
+  "Win2022x64_bios_CORE": {
```
akihikodaki, line 2: - It's 2019, not 2022.
- I think the `_CORE` suffix should be removed for consistency with the names of platforms that use it.

akihikodaki (changes requested):

lib/engines/hckinstall/iso.json
kostyanf14: in any case, this is just an example without real `product_key` this file does not work)

akihikodaki (approved):

lib/engines/hcktest/platforms/Win10_2004x86.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "Win10_2004x86",
+  "iso": "Win10_2004x86",
```
kostyanf14, line 3: As currently, we have separated client-specific options and platform options, and this ISO is a client iso only; update the name.

kostyanf14:
@akihikodaki Please review after latest changes

akihikodaki (approved):

PR #631, opened 2025-03-20
RHELMISC-10944: Fixes for run the same test several times

kostyanf14:


akihikodaki:
I suggested using `InstanceId` at https://github.com/HCK-CI/AutoHCK/pull/629#discussion_r2002914163; it may allow detecting retried tests whether the retry was triggered by AutoHCK or manually triggered. Is there something problematic when doing so?

kostyanf14:
@akihikodaki look at the code. I use `list_all_results` not tests to detect changes.

lib/engines/hcktest/tests.rb
```diff
@@ -208,7 +213,17 @@ def test_finished?(test)
     end
 
     def done_tests
-      @tests.select { |test| test_finished?(test) }
+      # When test is running more than ones HLK reports test['status'] = PASS/FAIL
+      # even when test is still running again. So we need to check test['executionstate']
+      # to be sure that test is really finished.
+      # Otherwise new_done function can report test as finished multiple times just with
+      # different executionstate.
+      @tests.select { |test| test_finished?(test) && test['executionstate'] == 'NotRunning' }
```
akihikodaki, line 221: I think this change is no longer necessary.

kostyanf14: we use `done_tests` for CI status. This change fixes displayed information.

akihikodaki: It will be inconsistent with `cnt_passed` and `cnt_failed` in `tests_stats`. Correct `cnt_passed` and `cnt_failed`, and derive `currentcount` by summing them.

lib/engines/hcktest/tools.rb
```diff
@@ -387,9 +393,11 @@ def apply_project_filters(project)
       end
     end
 
-    def zip_test_result_logs(test_id, target_key, machine, tag)
+    def zip_test_result_logs(test_id, target_key, machine, tag, index: 1, index_instance_id: false) # rubocop:disable Metrics/ParameterLists
```
akihikodaki, line 396: I guess you meant:
```suggestion
    def zip_test_result_logs(test_id, target_key, machine, tag, index: -1, index_instance_id: false) # rubocop:disable Metrics/ParameterLists
```

But I wonder if it is a good idea to have default arguments or even to name parameters here; it's possible to use three dots to make this simpler and remove the `Metrics/ParameterLists` warning:

```suggestion
    def zip_test_result_logs(...)
```

The default arguments or parameter names can be provided in rtoolsHCK instead.

kostyanf14: >even to name parameters here

yes, because Tools#zip_test_result_logs and RToolsHCK#zip_test_result_logs have partially different parameters. 

t.zip_test_result_logs(index, test_id, target_key, **tag**, machine, **tag**, index_instance_id:) - tag 2 times

lib/engines/hcktest/tests.rb
```diff
@@ -401,10 +437,12 @@ def apply_filters
     end
 
     def check_new_finished_tests
+      done, new_done = new_done_results
       return unless new_done.any?
 
       apply_filters
-      handle_finished_tests(new_done)
+      handle_finished_test_results(new_done)
+      @last_done_results = done
```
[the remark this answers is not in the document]
kostyanf14: but how will this help us? we pass new_done to handle_finished_test_results and do not call list_all_results again in this block.

lib/engines/hcktest/tests.rb
```diff
@@ -305,31 +320,41 @@ def update_summary_results_log
...
       @project.result_uploader.delete_file(r_name)
     end
 
     def all_tests_finished?
-      status_count('InQueue').zero? && current_test.nil?
+      status_count('InQueue').zero? &&
+        # TestResult.Status does not return Queued
+        # (if a test is scheduled or running it returns Running).
+        results_count('Running').zero? &&
+        current_test.nil?
```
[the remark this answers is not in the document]
kostyanf14: `tests_stats` and `all_tests_finished?` have a different behavior with the same name.

`tests_stats.InQueue` - how many tests still need to be run (In autohck queue)
`status_count('InQueue')` (all_tests_finished?) - how many tests are currently in the HLK controller queue. 

In this case, we need to know is there are any tests in the HLK controller queue or not

lib/engines/hcktest/tools.rb
kostyanf14: `index` fixed

kostyanf14:
Rspec error:
```
rspec './spec/kit_spec.rb[1:5]' # kit_spec Kit for ./lib/engines/hckinstall/kits/HLK1809.json can be downloaded
rspec './spec/kit_spec.rb[1:6]' # kit_spec Kit for ./lib/engines/hckinstall/kits/HLK1809server.json can be downloaded
```
Does not related to PR changes.

lib/engines/hcktest/tests.rb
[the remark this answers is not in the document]
akihikodaki: `results_count` and `done_test_results` both calls `list_all_results`. We can have one `list_all_results` call and store the returned value in an instance variable to reduce the number of `list_all_results` calls.

lib/engines/hcktest/tools.rb
akihikodaki: I wonder why we fiddle with parameters here instead of dealing with them at rtoolsHCK or `AutoHCK::Tests`. I particularly have an intuition that moving the logic to `AutoHCK::Tests` will result in simpler code.

lib/engines/hcktest/tools.rb
```diff
@@ -387,9 +393,11 @@ def apply_project_filters(project)
       end
     end
 
-    def zip_test_result_logs(test_id, target_key, machine, tag)
+    def zip_test_result_logs(...)
       retries ||= 0
-      ret = act_with_tools { _1.zip_test_result_logs(-1, test_id, target_key, tag, machine, tag) }
+      ret = act_with_tools do |t|
+        t.zip_test_result_logs(...)
+      end
```
akihikodaki, line 400: It is no longer necessary to make this multi-line.

lib/engines/hcktest/tests.rb
```diff
@@ -45,6 +45,10 @@ def list_tests(log: false)
       retry
     end
 
+    def load_tests_results
+      @tests_results = @tools.list_all_results(@target['key'], @client.name, @tag)
```
akihikodaki, line 49: Let's compute what `new_done_results` computes here to group the code modelling the change of test results into one method.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -45,6 +45,13 @@ def list_tests(log: false)
...
+
+      @last_done_results = done_test_results
+      @new_done_tests = @last_done_results - @last_handled_done_results
```
akihikodaki, line 52: To save instance variables:

```suggestion
      last_done_results = done_test_results
      @tests_results = @tools.list_all_results(@target['key'], @client.name, @tag)
      @new_done_tests = last_done_results - done_test_results
```

lib/engines/hcktest/tests.rb
```diff
@@ -305,31 +329,45 @@ def update_summary_results_log
...
+      r_name = "Failed_#{result_instance_id}_#{test_name}.zip"
+      @project.result_uploader.delete_file(r_name)
+      r_name = "Passed_#{result_instance_id}_#{test_name}.zip"
       @project.result_uploader.delete_file(r_name)
     end
 
     def all_tests_finished?
-      status_count('InQueue').zero? && current_test.nil?
+      # Update tests results to get the latest status of the tests
+      load_tests_results
```
akihikodaki, line 364: It is confusing to have a side-effect in a method suffixed with `?`.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
kostyanf14: `@new_done_tests = last_done_results - done_test_results` - NO

new_done_tests is  a diff between last_done_results and last_handled_done_results. We set last_handled_done_results in `def handle_finished_test_results` after handling new_done_tests. 

`@new_done_tests = last_done_results - done_test_results` is a diff between 2 calls of list_all_results, but this does check handle we it or not.

In case, we manually rerun several tests via VNC, we can have a raise condition new results exist but are not processed. That is the reason, why I set `last_handled_done_results` only after processing results.

akihikodaki: We can rename `load_tests_results` to reflect that it advances the iteration. `@test_results` at the beginning of this method will represent the previous iteration, and the updated value will represent the next iteration. We can model the state change of the loop this way.

kostyanf14: `update_tests_results`?

akihikodaki: That sounds good to me.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
```diff
@@ -28,7 +28,7 @@ def initialize(client, support, project, target, tools)
       @playlist = Playlist.new(client, project, target, tools, @client.kit)
       @tests = []
       @tests_extra = {}
-      @last_done = []
+      @tests_results = []
```
akihikodaki, line 31: Let's name this `@test_results`.

lib/engines/hcktest/tests.rb
```diff
@@ -45,6 +45,12 @@ def list_tests(log: false)
       retry
     end
 
+    def update_tests_results
+      last_done_results = done_test_results
+      @tests_results = @tools.list_all_results(@target['key'], @client.name, @tag)
+      @new_done_results = done_test_results - last_done_results
```
akihikodaki, line 51: For consistency:

```suggestion
      last_done_test_results = done_test_results
      @tests_results = @tools.list_all_results(@target['key'], @client.name, @tag)
      @new_done_test_results = done_test_results - last_done_test_results
```

lib/engines/hcktest/tests.rb
```diff
@@ -389,22 +433,19 @@ def reset_clients_to_ready_state
...
       @logger.info('Applying filters on finished tests')
       @tools.apply_project_filters(@tag)
       sleep APPLYING_FILTERS_INTERVAL
     end
 
-    def check_new_finished_tests
-      return unless new_done.any?
+    def check_new_finished_results
+      # Update tests results to get the latest status of the tests
+      update_tests_results
```
akihikodaki, line 444: Let's call this method at `handle_test_running` as it computes not only `@new_done_test_results` but also `@tests_results`, which affects more than the following `handle_finished_test_results` call.

lib/engines/hcktest/tests.rb
```diff
@@ -192,23 +198,36 @@ def status_count(status)
       @tests.count { |test| test['status'] == status }
     end
 
+    def results_count(status)
+      @tests_results.count { |test| test['status'] == status }
```
akihikodaki, line 202: The block parameter is named `test`, but it's a test result. `_1` will do anyway.

lib/engines/hcktest/tests.rb
```diff
@@ -305,31 +328,42 @@ def update_summary_results_log
...
       @project.result_uploader.delete_file(r_name)
     end
 
     def all_tests_finished?
-      status_count('InQueue').zero? && current_test.nil?
+      status_count('InQueue').zero? &&
+        # TestResult.Status does not return Queued
+        # (if a test is scheduled or running it returns Running).
+        results_count('Running').zero? &&
+        current_test.nil?
```
akihikodaki, line 366: What about checking only `@tests.any? { _1['executionstate'] != 'NotRunning' }`? According to `tests_stats_status_count`, it will check if there is any test not finished yet whether it is retried.

lib/engines/hcktest/tests.rb
```diff
@@ -354,30 +388,40 @@ def download_memory_dumps(l_tmp_path)
...
 
-        @project.github.update(tests_stats) if @project.github&.connected?
+    def test_for_result(result)
+      @tests.find { |test| test['name'] == result['name'] }
+    end
 
-        collect_memory_dumps(test)
+    def handle_finished_test_results(results)
+      # We need to list tests again to get the latest status of the tests
+      list_tests
```
akihikodaki, line 416: Better to group this and `update_tests_results` as they are matched.

akihikodaki (changes requested):

lib/engines/hcktest/tests.rb
kostyanf14: what about call `list_tests` in `update_tests_results`?

lib/engines/hcktest/tests.rb
kostyanf14: I don't see any reason to update this part.

lib/engines/hcktest/tests.rb
akihikodaki: That sounds good as long as `update_tests_results` is renamed accordingly.

lib/engines/hcktest/tests.rb
akihikodaki: It will improve consistency with `tests_stats_status_count` and also simplify the logic; now this method is comprised of three conditions but the condition I suggested may be sufficient. `status_count` and `results_count` methods and the comment in this method will be unnecessary.

lib/engines/hcktest/tests.rb
kostyanf14: update_tests_information?
update_tests_status?

lib/engines/hcktest/tests.rb
kostyanf14: You wrote inverted logic `@tests.any? { _1['executionstate'] != 'NotRunning' }`. `all_tests_finished?` = `@tests.all? { _1['executionstate'] == 'NotRunning' }` but anyway this does not work. 

HLK needs some time to prepare the client and before the client is ready (other "ready" than we check) test executionstate is NotRunning. And HLK special for this has two different fields `executionstate` and `status` but `status` changes to PASS/FAIL immediately after "main task" finished when multiple run of one test occurs. 

The same situation you can see in HLK Studio UI, just run one test two times and you can see that the status changes to PASS but one of the test results is still in progress. 
```
I, [2025-04-07T12:28:10.883307 #19723]  INFO -- : Adding to queue: run 1/2 [NdisStudio] OidRequestTest - OID_GEN_VENDOR_DRIVER_VERSION Verification (84490225-8259-4a2c-910c-1245f81740d7)
D, [2025-04-07T12:28:10.883474 #19723] DEBUG -- action/queue_test: test = 84490225-8259-4a2c-910c-1245f81740d7, target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu, sup is off, parameters is on, ipv6 is off
D, [2025-04-07T12:28:19.205322 #19723] DEBUG -- action/list_test_results: test_id = '84490225-8259-4a2c-910c-1245f81740d7', target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu
D, [2025-04-07T12:28:19.866904 #19723] DEBUG -- : Checking queued time for test id: 84490225-8259-4a2c-910c-1245f81740d7. Current test: 
D, [2025-04-07T12:28:19.867312 #19723] DEBUG -- : Test extra information: {"queued_at"=>#<DateTime: 2025-04-07T12:28:14+03:00 ((2460773j,34094s,205124406n),+10800s,2299161j)>}
D, [2025-04-07T12:28:19.867512 #19723] DEBUG -- action/list_tests: target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu, test_type is off, test_status is off, playlist is on
!!!!!!!!!!!!!!!!!! [NdisStudio] OidRequestTest - OID_GEN_VENDOR_DRIVER_VERSION Verification NotRunning InQueue
!!!!!!!!!!!!!!!!!! [NdisStudio] OidRequestTest - OID_GEN_VENDOR_ID Verification NotRun NotRunning
!!!!!!!!!!!!!!!!!! DF - InfVerif INF Verification NotRun NotRunning
I, [2025-04-07T12:28:23.165090 #19723]  INFO -- : Adding to queue: run 2/2 [NdisStudio] OidRequestTest - OID_GEN_VENDOR_DRIVER_VERSION Verification (84490225-8259-4a2c-910c-1245f81740d7)
D, [2025-04-07T12:28:23.165299 #19723] DEBUG -- action/queue_test: test = 84490225-8259-4a2c-910c-1245f81740d7, target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu, sup is off, parameters is on, ipv6 is off
D, [2025-04-07T12:28:28.464510 #19723] DEBUG -- action/list_test_results: test_id = '84490225-8259-4a2c-910c-1245f81740d7', target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu
D, [2025-04-07T12:28:28.892317 #19723] DEBUG -- : Checking queued time for test id: 84490225-8259-4a2c-910c-1245f81740d7. Current test: 
D, [2025-04-07T12:28:28.892483 #19723] DEBUG -- : Test extra information: {"queued_at"=>#<DateTime: 2025-04-07T12:28:23+03:00 ((2460773j,34103s,464319526n),+10800s,2299161j)>}
D, [2025-04-07T12:28:28.892613 #19723] DEBUG -- action/list_tests: target = PCI\VEN_1AF4&DEV_1041&SUBSYS_11001AF4&REV_01\4&C580B9D&0&0020, project = NetKVM-Win11_24H2x64_host_viommu, machine = CL1, pool = NetKVM-Win11_24H2x64_host_viommu, test_type is off, test_status is off, playlist is on
!!!!!!!!!!!!!!!!!! [NdisStudio] OidRequestTest - OID_GEN_VENDOR_DRIVER_VERSION Verification InQueue InQueue
!!!!!!!!!!!!!!!!!! [NdisStudio] OidRequestTest - OID_GEN_VENDOR_ID Verification NotRun NotRunning
!!!!!!!!!!!!!!!!!! DF - InfVerif INF Verification NotRun NotRunning
```

akihikodaki: What about: `@tests.all? { _1['status'] != 'InQueue' && _1['executionstate'] == 'NotRunning' }`
This will have the same advantages over the current code though the effectiveness is reduced.

Nothing changes with the current code when `_1['status'] == 'InQueue'`.
When `_1['status'] != 'InQueue'`, check for `_1['executionstate']` for the same reason as `tests_stats_status_count`. We can reuse the underlying logic of `tests_stats_status_count` this way.

This condition simplifies the logic by reducing the number of properties to check from three to two. More concretely, the code currently in this pull request checks the `status` and `executionstate` properties of `@tests` and the `status` property of `@test_results`. The condition I propose does not need to check the `status` property of `@test_results`.

lib/engines/hcktest/tests.rb
akihikodaki: What about merging this method into `list_tests` and naming it simply: `update_tests`?

kostyanf14: Rubocop ABC is too big. We can rename `list_tests` to `update_tests` but leave `update_tests_results` as a separate function and call it in `update_tests`. What about this?

akihikodaki: Rubocop didn't complain with the following:

```rb
    def list_tests(log: false)
      retries ||= 0
      last_done_test_results = done_test_results
      @test_results = @tools.list_all_results(@target['key'], @client.name, @tag)
      @new_done_test_results = done_test_results - last_done_test_results
      @tests = @playlist.list_tests(log)
    rescue Playlist::ListTestsError => e
      @logger.warn(e.message)
      @logger.info('Reconnecting tools...')
      @tools.reconnect
      raise unless (retries += 1) == 1 || verify_target

      @logger.info('Trying again to list tests')
      retry
    end
```

Calling `update_tests` from `update_tests_results` is awkward. `update_tests_results` is more descriptive and looks like it has a narrow scope; updating `@tests` is out of such scope.

lib/engines/hcktest/tests.rb
kostyanf14: `@tests.all? { _1['status'] != 'InQueue' && _1['executionstate'] == 'NotRunning' }` = `status_count('InQueue').zero? && current_test.nil?` but we also need to check `results_count` unfortunately. 

This is not a theoretical, but a practical issue. 
```
I, [2025-04-10T17:53:33.590489 #53308]  INFO -- : Adding to queue: run 2/2 DF - SimpleIO stress test with IO process termination (Reliability) (fc7e0020-bb7a-483f-b495-7e7fd0a40daa)
I, [2025-04-10T17:53:45.917292 #53308]  INFO -- : >>> Currently running: DF - SimpleIO stress test with IO process termination (Reliability) [00:02:00]
...
D, [2025-04-10T17:55:05.793822 #53308] DEBUG -- : Checking queued time for test id: fc7e0020-bb7a-483f-b495-7e7fd0a40daa. Current test: {"name"=>"DF - SimpleIO stress test with IO process termination (Reliability)", "id"=>"fc7e0020-bb7a-483f-b495-7e7fd0a40daa", "testtype"=>"Automated", "estimatedruntime"=>"00:02:00", "requiresspecialconfiguration"=>"True", "requiressupplementalcontent"=>"False", "scheduleoptions"=>["ScheduleOnAnyTarget", "ConsolidateScheduleAcrossTargets"], "status"=>"Passed", "executionstate"=>"Running", "url"=>"https://docs.microsoft.com/en-us/windows-hardware/test/hlk/testref/fc7e0020-bb7a-483f-b495-7e7fd0a40daa", "run_count"=>2, "duration"=>120.0}
D, [2025-04-10T17:55:05.793899 #53308] DEBUG -- : Test extra information: {"queued_at"=>#<DateTime: 2025-04-10T17:53:33+03:00 ((2460776j,53613s,738266861n),+10800s,2299161j)>, "started_at"=>#<DateTime: 2025-04-10T17:53:45+03:00 ((2460776j,53625s,917308428n),+10800s,2299161j)>, "status"=>nil}
D, [2025-04-10T17:55:19.293790 #53308] DEBUG -- : Checking queued time for test id: fc7e0020-bb7a-483f-b495-7e7fd0a40daa. Current test: 
!!!!!! NO TEST RESULT SAVING
I, [2025-04-10T17:55:29.294778 #53308]  INFO -- : Exiting HLKClient CL2
```

Test status changed to `Passed` just after the test `executionstate` moved from `InQueue` to `Running` -  known HLK behavior number 1. 
When the test runs several times, the test `executionstate` moves from `Running` to `NotRunning` just after the main part of the test is finished even if the cleanup stage is still running  -  known HLK behavior number 2. 

As a result @new_done_test_results = [], because `test result status` is not `PASS/FAIL` yet. 
I agree that this HLK behavior is very not clean, but this practically happened.

lib/engines/hcktest/tests.rb
kostyanf14: This variant broke the scenario with manual rerun tests via VNC.

1. I re-run tests via VNC
2. AutoHCK was at the end of the `handle_test_running` loop and exited it because all test was finished
3. AutoHCK queued the next test and called `list_tests`. But with this implementation, `list_tests` updated `@new_done_test_results` at the point when AutoHCK does not check it. As a result, in logs, I missed 2 manual runs. 

To merge `update_tests_results` into `list_tests` we need back the old logic where `last_done_test_results` is set only after `handle_finished_test_results` to make sure that we processed `@new_done_test_results`.

We must call `list_tests` after `queue_test` outside of the `handle_test_running` loop, to update the `@tests` list to make `all_tests_finished? = false`

akihikodaki: It implies we have an extra `list_tests` call between two `handle_test_running` calls. We can remove the extra `list_tests` call by checking `all_tests_finished?` after calling `list_tests` *in loop*.

lib/engines/hcktest/tests.rb
akihikodaki: I understand. It explains the situation well.

lib/engines/hcktest/tests.rb
```diff
@@ -192,23 +201,36 @@ def status_count(status)
...
     end
 
-    def done_tests
-      @tests.select { |test| test_finished?(test) }
+    def tests_stats_status_count(status)
+      # When test is running more than once HLK reports test['status'] = PASS/FAIL
+      # even when test is still running again. So we need to check test['executionstate']
+      # to be sure that test is really finished.
+      # Otherwise new_done function can report test as finished multiple times just with
+      # different executionstate.
```
akihikodaki, line 228: `new_done` no longer exists.

lib/engines/hcktest/tests.rb
```diff
@@ -305,31 +331,42 @@ def update_summary_results_log
...
+      r_name = "Passed_#{result_instance_id}_#{test_name}.zip"
       @project.result_uploader.delete_file(r_name)
     end
 
     def all_tests_finished?
-      status_count('InQueue').zero? && current_test.nil?
+      status_count('InQueue').zero? &&
+        # TestResult.Status does not return Queued
+        # (if a test is scheduled or running it returns Running).
+        results_count('Running').zero? &&
```
akihikodaki, line 368: `status_count` and `results_count` may not be worthwhile to have methods; they are trivial and currently called only in this method. We can also simplify them a bit by using `Array#any?` instead of `Array#count`.

akihikodaki (changes requested):

akihikodaki:
> @akihikodaki Based on [#631 (comment)](https://github.com/HCK-CI/AutoHCK/pull/631#discussion_r2038072248) and [#631 (comment)](https://github.com/HCK-CI/AutoHCK/pull/631#discussion_r2037802694). Currently, there is no critical issue, maybe some potential source cleanup but every try to cleanup causes some other issue / broken scenario that works now.
> 
> Manual rerun during HLK run is not a very often scenario, but we use it. Rerun test several times is used by developers to capture some data/logs when an error is not stable enough to reproduce.
> 
> So, maybe it is time to stop refactoring/improvements and merge this?

The discussion about `all_tests_finished?` showed it's fine, but one about `update_tests_status` implies a logic error; `all_tests_finished?` in `handle_test_running` uses both `@tests` and `@test_results`, but `run` only updates `@tests` before calling `handle_test_running`. This discrepancy of `@tests` and `@test_results` should be fixed. I also hope it makes naming easier though that's not a blocker.

lib/engines/hcktest/tests.rb
kostyanf14: Sorry, can you explain in more detail.

lib/engines/hcktest/tests.rb
kostyanf14: fixed

lib/engines/hcktest/tests.rb
akihikodaki: Simply merging `update_tests_results` into `list_tests` and rename it `update_tests` will break the procedure to start a test run.
1. `run` queues a test run.
2. `run` calls `update_tests`.
3. `run` calls `handle_test_running`.
4. `handle_test_running` calls `all_tests_finished?`, using the result of the last `update_tests` call.
5. `handle_test_running` calls `update_tests`.

Here, `update_tests` calls at 2 and 5 are duplicates. This duplication causes an incorrect update of `@new_done_test_results` that you described in the last comment.
However, even without merging, the duplication is inefficient as the `list_tests` RPC calls are duplicated. This duplication is better resolved anyway by removing 2 and reordering 4 and 5.

akihikodaki (approved):

kostyanf14:
@YanVugenfirer DON'T merge, I want to make more tests before

PR #635, opened 2025-03-27
RHELMISC-10658: Implement log upload to external FS

kostyanf14:
Can be used to mount NFS storage and upload logs to it.

~~This PR is a draft due to an issue with AutoHCK.log uploading.
As a result uploaders are created after the Project, we close them before closing the project. 
It caused the problem, that AutoHCK.log must be uploaded after result uploaders were closed.
We did not see any errors before, because `def close` is empty in Dropbox/S3 uploaders but even logically this is a problem.~~

```
I, [2025-03-27T20:45:58.530397 #58880]  INFO -- CmdRun PID:73517: Command finished with code 0
I, [2025-03-27T20:45:58.541493 #58880]  INFO -- CmdRun PID:73521: Run command: gio mount 'smb://WORKGROUP;iadmin@10.7.2.11/kostyanf14' --unmount
I, [2025-03-27T20:45:58.557758 #58880]  INFO -- CmdRun PID:73521: stderr: 
I, [2025-03-27T20:45:58.558214 #58880]  INFO -- CmdRun PID:73521: stderr: (gio mount:73521): GVFS-WARNING **: 20:45:58.557: The peer-to-peer connection failed: Exhausted all available authentication mechanisms (tried: EXTERNAL) (available: EXTERNAL). Falling back to the session bus. Your application is probably missing --filesystem=xdg-run/gvfsd privileges.
I, [2025-03-27T20:45:58.561212 #58880]  INFO -- CmdRun PID:73521: Command finished with code 0
D, [2025-03-27T20:45:58.561406 #58880] DEBUG -- : Closing AutoHCK project
W, [2025-03-27T20:45:58.561955 #58880]  WARN -- : FS upload_file error: No such file or directory @ rb_sysopen - /run/user/1000/gvfs/smb-share:domain=WORKGROUP,server=10.7.2.11,share=kostyanf14,user=iadmin//auto_hck/kostyanf14/test1/NetKVM-Win11_24H2x64_host_viommu/2025_03_27_19_39_29/AutoHCK.log (Errno::ENOENT)
```

~~As an idea, we can initialize result uploaders before to project, then configure them to load data from the project and then we will close it after the project.~~
@akihikodaki @Jedoku Any comments/suggestions?

### DIFF
--- a/lib/resultuploaders/fs_copy/fs_copy.rb
+++ b/lib/resultuploaders/fs_copy/fs_copy.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+module AutoHCK
+  class FsCopy
+    CONFIG_JSON = 'lib/resultuploaders/fs_copy/fs_copy.json'
+
+    include Helper
+
+    attr_reader :url
+
+    def initialize(project)
+      @logger = project.logger
+
+      @config = Json.read_json(CONFIG_JSON, @logger)
+
+      @host_path = @config['host_path']
+      base_url = @config['base_url']
+
+      work_path = "auto_hck/#{project.config['repository']}/#{project.engine_tag}/#{project.timestamp}"
+
+      @path = "#{@host_path}/#{work_path}"
+      @url = "#{base_url}/#{work_path}" unless base_url.nil?
+    end
+
+    def html_url; end
+
+    def handle_exceptions(where)
+      yield
+    rescue StandardError => e
+      @logger.warn("FS #{where} error: #{e.detailed_message}")
+      false
+    end
+
+    # These methods are intentionally left blank to maintain the interface,
+    # as FsCopy does not require this function. It prevents external calls
+    # from causing errors while adhering to the expected API structure.
+    def ask_token; end
+
+    def connect
+      File.exist?(@host_path)
+    end
+
+    def create_project_folder
+      handle_exceptions(__method__) do
+        FileUtils.mkdir_p(@path)
+        @logger.info("FS project folder created: #{@url}")
+      end
+    end
+
+    def upload_file(l_path, r_name)
+      handle_exceptions(__method__) do
+        r_path = "#{@path}/#{r_name}"
+        FileUtils.copy(l_path, r_path)
+      end
+    end
+
+    def update_file_content(content, r_name)
+      handle_exceptions(__method__) do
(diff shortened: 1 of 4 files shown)


akihikodaki (commented):
What about mounting before running AutoHCK and unmounting after? We can save the code to handle `connect_commands` and `close_commands` by doing so.

kostyanf14:
> What about mounting before running AutoHCK and unmounting after? We can save the code to handle `connect_commands` and `close_commands` by doing so.

In this case, we need an external script to do this. We need this for RH gating tests, and the current QE solution mounts NFS after starting. It even detects which NFS is better to use based on ping.

akihikodaki:
> > What about mounting before running AutoHCK and unmounting after? We can save the code to handle `connect_commands` and `close_commands` by doing so.
> 
> In this case, we need an external script to do this. We need this for RH gating tests, and the current QE solution mounts NFS after starting. It even detects which NFS is better to use based on ping.

I'm thinking of a scenario where there is a shell script or CI configuration file that contains a command line to invoke AutoHCK. Adding commands to mount there sounds simpler.

akihikodaki (commented):
I wonder why `project` is now passed to `connect`

kostyanf14:
> I wonder why `project` is now passed to `connect`

Because `project` is not ready when `ResultUploader#initialize` is called.

akihikodaki:
> > I wonder why `project` is now passed to `connect`
> 
> Because `project` is not ready when `ResultUploader#initialize` is called.

`ResultUploader#initialize` still takes `project` as its parameter. The reason why e.g., `S3#initialize` does not do the same needs to be explained.

lib/resultuploaders/fs_copy/fs_copy.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      @config = Json.read_json(CONFIG_JSON, @logger)
+
+      @host_path = @config['host_path']
+      base_url = @config['base_url']
+
+      work_path = "auto_hck/#{project.config['repository']}/#{project.engine_tag}/#{project.timestamp}"
+
+      @path = "#{@host_path}/#{work_path}"
+      @url = nil
```
akihikodaki, line 22:
```suggestion
```

akihikodaki (changes requested):

akihikodaki (approved):

PR #636, opened 2025-03-28
RHELMISC-10661: JUnit: Implement JUnit XML builder (initial commit)

kostyanf14:
@akihikodaki 

I didn't find any gem to generate JUnit XML on Ruby. Maybe export this implementation as a separate gem. What do you think?


akihikodaki (changes requested):
I think it's better to have classes decoupled from JUnit that represent the results. Such classes may have e.g., a method to count failed tests and be shared with HTML and YAML reporters.

Nokogiri is a complicated library with native code. Heredocs and XML encodings should be sufficient for our use case: https://docs.ruby-lang.org/en/3.4/encodings_rdoc.html#label-Encoding+Options

kostyanf14:
> think it's better to have classes decoupled from JUnit that represent the results. Such classes may have e.g., a method to count failed tests and be shared with HTML and YAML reporters.

in general, I agree, but in JUnit::TestSuite we need to calculate a count of test cases. We can't just write `failed=8` without adding 8 failed tests. If we will calculate this data outside of the JUnit context, theoretically we can have incomplete XML.

> Nokogiri is a complicated library with native code. Heredocs and XML encodings should be sufficient for our use case: https://docs.ruby-lang.org/en/3.4/encodings_rdoc.html#label-Encoding+Options

Do you suggest generating XML manually with Heredocs?

akihikodaki:
> > think it's better to have classes decoupled from JUnit that represent the results. Such classes may have e.g., a method to count failed tests and be shared with HTML and YAML reporters.
> 
> in general, I agree, but in JUnit::TestSuite we need to calculate a count of test cases. We can't just write `failed=8` without adding 8 failed tests. If we will calculate this data outside of the JUnit context, theoretically we can have incomplete XML.
> 
> > Nokogiri is a complicated library with native code. Heredocs and XML encodings should be sufficient for our use case: https://docs.ruby-lang.org/en/3.4/encodings_rdoc.html#label-Encoding+Options
> 
> Do you suggest generating XML manually with Heredocs?

Yes. You can call a method to count failed tests when doing so.

YanVugenfirer:
> I think it's better to have classes decoupled from JUnit that represent the results. Such classes may have e.g., a method to count failed tests and be shared with HTML and YAML reporters.
> 
> Nokogiri is a complicated library with native code. Heredocs and XML encodings should be sufficient for our use case: https://docs.ruby-lang.org/en/3.4/encodings_rdoc.html#label-Encoding+Options

If it works on ARM, I don’t see a reason not to use it.

kostyanf14:
@akihikodaki 

We can use https://github.com/rails/builder instead

akihikodaki:
> @akihikodaki
> 
> We can use https://github.com/rails/builder instead

It's excessive for our use case. It's useful when encoding one data type for multiple interfaces (e.g., when using a XML-based API with multiple endpoints), but our use case is different; in our use case, the encoding of `testcase` element and other elements are used only to generate one-specific XML format and we have no motivation for reusing them. Hence, we only need one monolithic template code for JUnit XML, and it is not necessary to introduce a library for that.

kostyanf14:
> It's excessive for our use case

Why? 

I don't like the idea of manually generating any structured data especially when some values should be escaped. Yes, XML has a simple format in general, but we need to be careful with all data that passes as an attribute to escape XML-specific symbols.  

@YanVugenfirer What do you think?

akihikodaki:
> > It's excessive for our use case
> 
> Why?
> 
> I don't like the idea of manually generating any structured data especially when some values should be escaped. Yes, XML has a simple format in general, but we need to be careful with all data that passes as an attribute to escape XML-specific symbols.
> 
> @YanVugenfirer What do you think?

I thought of using ERB for automatic escaping, but it may be more complicated than having heredocs and a few simple methods that write attributes and texts with escaping.

By the way, texts in existing ERB templates apparently have not been properly escaped. The ERB implementation we use is the one bundled in Ruby, but it does not automatically escape texts. Rails uses Erubi, which automatically escapes texts.

YanVugenfirer:
I don’t think manual generation of XML files is a good idea

kostyanf14:
@akihikodaki, this revision is just to sync the implementation direction.

kostyanf14:
@akihikodaki This is the initial implementation that provides only HLK tests as JUnit test cases

lib/junit.rb
```diff
@@ -0,0 +1,53 @@
...
+        'host' => @project.setup_manager&.host_info,
+        'results_url' => @project.result_uploader&.url,
+        'results_html' => @project.result_uploader&.html_url
+      }
+    end
+
+    def report_data
+      {
+        'tag' => @project.engine_tag,
+        'auto_hck_log' => @project.string_log.string.encode('ASCII', xml: :text),
```
akihikodaki, line 34:
```suggestion
        'auto_hck_log' => @project.string_log.string.encode('UTF-8', xml: :text),
```

But I think it's better to fix the entire problem class by introducing Erubi.

lib/templates/junit.xml.erb
```diff
@@ -0,0 +1,41 @@
...
+        skipped="<%= engine_test_steps.count { _1.is_skipped }%>"
+        time="<%= engine_test_steps.sum { _1.execution_time } %>"
+        timestamp="<%= timestamp %>">
+        <properties>
+            <% run_properties.each do |name, value| %>
+                <property name="<%= name %>" value="<%= value %>" />
+            <% end %>
+        </properties>
+
+        <% unless auto_hck_log.nil? %>
```
akihikodaki, line 24: I don't think `auto_hck_log` can be `nil`.

lib/junit.rb
```diff
@@ -0,0 +1,53 @@
...
+        'timestamp' => Time.strptime(@project.timestamp, '%Y_%m_%d_%H_%M_%S').iso8601,
+        'run_properties' => run_properties,
+        'engine_test_steps' => @project.engine&.test_steps&.select do |step|
```
akihikodaki, line 37: `&.` implies this can be `nil` but the template assumes it's not `nil`.

akihikodaki (changes requested):

lib/junit.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+
+    def generate_junit_report(file_path)
+      @logger.info('Generating JUnit report')
+      set_report_data
+      @logger.debug("Writing JUnit report to #{file_path}")
+
+      # rubocop:disable Security/Eval
+      File.write(file_path, eval(@template.src))
```
akihikodaki, line 53: Calling `eval` here is a bit inefficient as it requires parsing generated Ruby code each time this method is called. Instead of keeping an `Erubi::Engine`, we can keep the evaluated code instead by defining a method with `eval`. Please refer to Rails for example: https://github.com/rails/rails/commit/5026abaeb17fb8c430219f0df0c5c44e5c57e173

lib/project.rb
```diff
@@ -58,21 +59,19 @@ def log_exception(exception, level)
...
       @logfile_name = logfile_name
       @logfile_path = "#{workspace_path}/#{@logfile_name}"
-      IO.copy_stream @temp_pre_logger, @logfile_path
-      @pre_logger.close
-      @logger.remove_logger(@pre_logger)
-      @pre_logger = MonoLogger.new(@logfile_path)
-      @logger.add_logger(@pre_logger)
+      IO.copy_stream @string_log, @logfile_path
+      file_logger = MonoLogger.new(@logfile_path)
+      @logger.add_logger(file_logger)
```
akihikodaki, line 74: Simpler:

```suggestion
      @logger.add_logger(MonoLogger.new(@logfile_path))
```

akihikodaki (changes requested):
The code that uses Erubi is better to extract:
- to reuse for HTML generation in the future and
- to encapsulate `escape` and `eval` for correct usage.

The code else `AutoHCK::Junit` looks good to me.

lib/junit.rb
```diff
@@ -0,0 +1,66 @@
...
+      @logger = project.logger
+
+      template = Erubi::Engine.new(File.read('lib/templates/junit.xml.erb'), escape: true)
+      source = <<-RUBY
+        def build_xml
+          #{template.src}
+        end
+      RUBY
+
+      JUnit.module_eval(source)
```
akihikodaki, line 16: This should be moved out of the `initialize` method as it initializes the class instead of an instance.

akihikodaki (changes requested):
Sorry, I missed one remaining problem with template loading.

lib/junit.rb
```diff
@@ -0,0 +1,71 @@
...
+    # rubocop:disable Security/Eval
+    eval <<-RUBY, binding, __FILE__, __LINE__ + 1
+      # def build_xml
+      #   Erubi template compiled source code
+      # end
+
+      def build_xml
+        #{template.src}
+      end
+    RUBY
```
akihikodaki, line 16: I guess you intended to make a stack trace generated in the template point to line 10, but it will point to a line after line 14.

A line number of an Erubi's output apparently matches the template source code, so I suggest the following:
```suggestion
    path = 'lib/templates/junit.xml.erb'
    binding.eval "def build_xml; #{template.src}; end", path
```

By the way, looking at Erubi's output, I came up with a trivial suggestion for the template so I leave the comments for it too.

akihikodaki (changes requested):

lib/junit.rb
kostyanf14: Oneline is bad due to rubocop offence Style/DocumentDynamicEvalDefinition, this is the reason for the comment.

kostyanf14:
```
    template = Erubi::Engine.new(File.read('lib/templates/junit.xml.erb'), escape: true)
    path = 'lib/templates/junit.xml.erb'

    # rubocop:disable Security/Eval
    binding.eval <<-RUBY, path
      # def build_xml
      #   Erubi template compiled source code
      # end

      def build_xml
        #{template.src}
      end
    RUBY
    # rubocop:enable Security/Eval
```

akihikodaki: The multi-line heredoc breaks the line number. Considering the code is so trivial, I think it's OK to disable Style/DocumentDynamicEvalDefinition here.

kostyanf14: Fixed

lib/junit.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module AutoHCK
+  class JUnit
+    template = Erubi::Engine.new(File.read('lib/templates/junit.xml.erb'), escape: true)
+    path = 'lib/templates/junit.xml.erb'
```
akihikodaki, line 6:
```suggestion
    path = 'lib/templates/junit.xml.erb'
    template = Erubi::Engine.new(File.read(path), escape: true)
```

akihikodaki (changes requested):

akihikodaki (approved):

PR #650, opened 2025-04-14
RHELMISC-11270: QemuHCK: Attach control network device to pcie-root-port

kostyanf14:
This allows to use NetKVM as a control network device with IOMMU enabled.

virtio-pci swithes to legacy mode, when device attached to pcie bus instead of root port.

This brake backward compatibility as a PCI topology changes and Windows reconfigure devices.


kostyanf14:
CI4CI fail expectedly

PR #705, opened 2025-06-27
GitHub status improvments

kostyanf14:


YanVugenfirer (approved):

akihikodaki (changes requested):
The usage of `Proc` is questionable. It makes the timing when the URL is computed dynamic and harder to figure out.

I also suspect you are making this conversion because some uploader will not be able to return a URL when `Github.new` is called. However, passing `url_proc` to `Github.new` implies the `Proc` is immediately available, which is contradictory.

A better approach would be to pass the URL (instead of using `Proc`) to the `Github` instance when it's available, which clarifies when the `Github` instance can expect the availability of URL. In any case, clarifying the motivation for this conversion will help.

I also have a nitpick: the comment saying `Raised on errors in the 500-599 range` looks redundant to me. Perhaps leaving a comment telling a meaning of the magic "5xx" number may make sense, but the inverse (a comment noting a magic number) may not.

PR #723, opened 2025-07-31
RHELMISC-15636: Generate PCI bus for each device that require it

kostyanf14:


YanVugenfirer:
Just to double check - how this change will behave with automatic installation of images? 
Is there a scenario where the bus changes after installation for the device, or for now only devices under test will be affected?

YanVugenfirer (approved):

PR #799, opened 2025-11-14
Set a configurable client/server installation password

menli820:
Set a configurable HLK-client and HLK-server
installation password.

depend on:  https://github.com/HCK-CI/HLK-Setup-Scripts/pull/27 
https://github.com/HCK-CI/AutoHCK-Installer/pull/44

### DIFF
--- a/lib/engines/hckinstall/hckinstall.rb
+++ b/lib/engines/hckinstall/hckinstall.rb
@@ -338,7 +338,8 @@ module AutoHCK
         '@WINDOWS_IMAGE_NAME@' => @studio_iso_info['studio']['windows_image_names'],
         '@PRODUCT_KEY@' => product_key,
         '@PRODUCT_KEY_XML@' => product_key_xml(product_key),
-        '@HOST_TYPE@' => 'studio'
+        '@HOST_TYPE@' => 'studio',
+        '@DEFAULT_PASSWORD@' => @project.config['studio_password']
       }
       @answer_files.each do |file|
         file_gsub(build_studio_answer_file_path(file),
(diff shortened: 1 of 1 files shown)


lib/engines/hckinstall/hckinstall.rb
```diff
@@ -241,6 +241,35 @@ def run_second(client:)
       end
     end
 
+    def read_password_from_args
+      args_file = "#{@hck_setup_scripts_path}/args.ps1.example"
```
kostyanf14: No need to parse `args.ps1.example`. This is just an example and in this case we still duplicate knowlage about default password. Let's leave it only in `config.json`.

lib/engines/hckinstall/hckinstall.rb
```diff
@@ -338,7 +367,8 @@ def prepare_studio_drives
         '@WINDOWS_IMAGE_NAME@' => @studio_iso_info['studio']['windows_image_names'],
         '@PRODUCT_KEY@' => product_key,
         '@PRODUCT_KEY_XML@' => product_key_xml(product_key),
-        '@HOST_TYPE@' => 'studio'
+        '@HOST_TYPE@' => 'studio',
+        '@DEFAULT_PASSWORD@' => read_password_from_args
```
kostyanf14: Based on comment below just use `@project.config['studio_password']` there.

config.json
```diff
@@ -5,7 +5,7 @@
     "winrm_port": 5985,
     "repository": "virtio-win/kvm-guest-drivers-windows",
     "studio_username": "Administrator",
-    "studio_password": "Qum5net.",
+    "studio_password": "Assentor01",
```
kostyanf14: I prefered don't touch value in `config.json` because we already have a lot of deployments that use default password. For new deployments lets set it in `override.json`. Maybe add question in AutoHCK-Installer.

kostyanf14 (approved):

PR #912, opened 2026-03-29
Add UI executor to bridge WinRM commands to the interactive desktop session

elizashurov:
WinRM commands run in Session 0 which cannot interact with the UI. This commit adds a PowerShell agent and Ruby executor that bridge WinRM to the interactive desktop session via a scheduled task with Interactive logon type and a filesystem-based command queue.

Enables running UI-dependent HLK tests that were previously rejected because they required interactive session access.

### DIFF
--- a/lib/auxiliary/ui_agent/ui_executor.rb
+++ b/lib/auxiliary/ui_agent/ui_executor.rb
@@ -1,0 +1,278 @@
+# typed: false
+# frozen_string_literal: true
+
+module AutoHCK
+  # Bridges WinRM (Session 0) to the interactive desktop (Session 1).
+  #
+  # WinRM runs in a non-interactive service session and cannot touch
+  # the UI. This class deploys a PowerShell agent into the user's
+  # desktop session via a scheduled task with LogonType Interactive,
+  # then communicates through a filesystem queue:
+  #
+  #   1. Host writes a command to  C:\UIAgent\queue\<id>.ps1
+  #   2. Agent executes it and writes C:\UIAgent\queue\<id>.json
+  #   3. Host polls for the .json result and cleans up
+  #
+  # If the agent dies (e.g. after a VM reboot that changes the
+  # interactive user), it is automatically redeployed on next use.
+  class UIExecutor
+    AGENT_DIR = 'C:\\UIAgent'
+    QUEUE_DIR = "#{AGENT_DIR}\\queue".freeze
+    AGENT_PATH = "#{AGENT_DIR}\\auto_hck_ui_agent.ps1".freeze
+    READY_PATH = "#{AGENT_DIR}\\agent.ready".freeze
+    LOG_PATH = "#{AGENT_DIR}\\agent.log".freeze
+    TASK_NAME = 'AutoHCK_UIAgent'
+    LOCAL_AGENT = File.join(__dir__, 'auto_hck_ui_agent.ps1')
+
+    DEFAULT_TIMEOUT = 120
+    POLL_INTERVAL = 2
+    DEPLOY_VERIFY_RETRIES = 10
+    DEPLOY_VERIFY_INTERVAL = 3
+
+    class DeployError < AutoHCKError; end
+    class CommandTimeoutError < AutoHCKError; end
+    class CommandError < AutoHCKError; end
+
+    def initialize(tools, machine, logger, username)
+      @tools = tools
+      @machine = machine
+      @logger = logger
+      @username = username
+      @deployed = false
+    end
+
+    # Deploys the agent if not already running
+    def deploy
+      if agent_alive?
+        @logger.info("UI agent already running on #{@machine}, skipping deploy")
+        @deployed = true
+        return
+      end
+
+      @logger.info("Deploying UI agent on #{@machine}")
+      create_directories
+      upload_agent
+      register_startup_task
+      remove_stale_ready_marker
+      start_agent
+      verify_agent_running
+      @deployed = true
+      @logger.info("UI agent deployed and running on #{@machine}")
+    rescue StandardError => e
+      raise DeployError, "Failed to deploy UI agent on #{@machine}: #{e.message}"
+    end
+
+    # Executes a PowerShell command in the interactive session.
+    # Deploys the agent on first call if not already running.
+    # Checks agent liveness before each command and redeploys if dead.
+    # If the agent dies during a command, the timeout catches it and
+    # redeploys once before raising.
+    def run(command, timeout: DEFAULT_TIMEOUT)
+      deploy unless @deployed
+
+      unless agent_alive?
+        @logger.warn("UI agent appears dead on #{@machine}, redeploying before command")
+        @deployed = false
+        deploy
+      end
+
+      execute_command(command, timeout)
+    rescue CommandTimeoutError
+      raise if agent_alive?
+
+      @logger.warn("UI agent died during command on #{@machine}, redeploying and retrying")
+      @deployed = false
+      deploy
+      execute_command(command, timeout)
+    end
+
+    def teardown
+      @logger.info("Tearing down UI agent on #{@machine}")
+      @tools.run_on_machine(
+        @machine,
+        'Stop UI agent task',
+        "schtasks /Delete /TN #{TASK_NAME} /F"
+      )
+      @deployed = false
+    rescue StandardError => e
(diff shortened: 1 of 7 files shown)


elizashurov:
Tested on a Win11_25H2x64 client with the viogpu driver, with AutoIt installed on the  VM and AutoIt .au3 watcher scripts handling UI dialogs.
HLK tests that previously required manual UI interaction now pass automatically:
* `Test for EDID Requirements (Manual)` - Passed
* `WDDM HPD Notification Test (Manual)` - Passed

Screenshot From 2026-03-29 12-00-18

lib/models/command_info.rb
```diff
@@ -51,6 +51,7 @@ class CommandInfo < T::Struct
       const :desc, String
       const :host_run, T.nilable(String)
       const :guest_run, T.nilable(String)
+      const :guest_ui_run, T.nilable(String)
```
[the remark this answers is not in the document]
kostyanf14: no need

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+    # via pre/post_test_commands.
+    def upload_watchers
+      return unless File.directory?(LOCAL_WATCHERS)
+
+      Dir.glob(File.join(LOCAL_WATCHERS, '*.au3')).each do |file|
+        remote = "#{WATCHERS_DIR}\\#{File.basename(file)}"
+        @logger.info("Uploading watcher #{File.basename(file)} to #{@machine}")
+        @tools.upload_to_machine(@machine, file, remote)
+      end
+    end
```
kostyanf14: No need this. Let's use the files_action mechanism when configuring tests. Don't need to add some specific logic only for one type of task. We have general API that can cover this situation.

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+        'Detect interactive session user',
+        '(Get-WmiObject -Class Win32_ComputerSystem).UserName'
+      ).to_s.strip
+      @logger.debug("Interactive session user on #{@machine}: '#{user}'")
+      user.empty? ? 'Administrator' : user.split('\\').last
+    end
+
+    # Interactive logon type places the process in Session 1 (desktop).
+    def register_startup_task
+      user = interactive_user
```
kostyanf14: Why do we need both 'register_startup_task' (via ScheduledTask) and 'register_agent_autostart' (via HKLM registry)?

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+
+    def log_result(id, result)
+      exit_code = result['exit_code']
+      if exit_code.zero?
+        @logger.info("UI command #{id} succeeded on #{@machine} (exit_code=0)")
+      else
+        @logger.warn("UI command #{id} failed on #{@machine} (exit_code=#{exit_code})")
+      end
+
+      @logger.info("UI command #{id} stdout: #{result['stdout']}") unless result['stdout'].to_s.empty?
```
kostyanf14: debug there is ok

lib/engines/hcktest/tests.rb
```diff
@@ -290,6 +314,7 @@ def run_file_actions(files_actions, replacement)
     def run_test_commands(test, type, replacement)
       select_test_config(test.name, type).each do |command|
         run_guest_test_command(command, replacement)
+        run_guest_ui_test_command(command, replacement)
         run_host_test_command(command)
 
         run_file_actions(command.files_action, replacement)
```
kostyanf14, line 319: Please update `docs/Test-config-behavior.md` especially the "Test Configuration Execution Order" section. Also, describe commands' lifetime and whether it is sync or async.

lib/models/command_info.rb
kostyanf14: Maybe better to use a boolean instead. This flag will describe how to run the command (via WinRM session 0 or via GUI executor). Default: false for backward compatibility. 

This allows you to simplify some code where you are looking for a command.

@YanVugenfirer what do you think?

lib/setupmanagers/hckclient.rb
```diff
@@ -77,17 +77,33 @@ def set_machine_ready
...
+          @logger.info("Running command (#{desc}) on client #{@name}")
+          updated_command = @replacement_map.create_cmd(command.guest_run)
+          @logger.debug("Running command after replacement (#{desc}) on client #{@name}: #{updated_command}")
+          @tools.run_on_machine(@name, desc, updated_command)
+        end
+
+        if command.guest_ui_run
+          @logger.info("Running UI command (#{desc}) on client #{@name}")
+          updated_command = @replacement_map.replace(command.guest_ui_run)
+          @logger.debug("Running UI command after replacement (#{desc}) on client #{@name}: #{updated_command}")
```
kostyanf14: Better to say: "Running command in an interactive session", because the command can be a CLI command, but requires Session 1 access.

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+    # Checks ready marker, then process command line, then task status
+    # to cover both initial launch and post-reboot restart via Run key.
+    def agent_alive?
+      return false unless @tools.exists_on_machine?(@machine, READY_PATH)
+
+      check = @tools.run_on_machine(
+        @machine,
+        'Check UI agent process',
+        "Get-WmiObject Win32_Process -Filter \"Name='powershell.exe'\" | " \
+        "Where-Object { $_.CommandLine -like '*agent.ps1*' } | " \
```
kostyanf14: 'agent' is too generic, ok for now, but let's make it more specific for the future. 
`auto_hck_ui_agent`

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+        "Select-Object -First 1 | ForEach-Object { 'agent_running' }"
+      )
+      return true if check.to_s.include?('agent_running')
+
+      status = @tools.run_on_machine(
+        @machine,
+        'Check UI agent task status',
+        "schtasks /Query /TN #{TASK_NAME} /FO CSV /NH"
+      )
+      status.to_s.include?('Running')
```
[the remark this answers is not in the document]
kostyanf14: @elizashurov Copilot can be correct there

lib/auxiliary/ui_agent/ui_executor.rb
```diff
@@ -0,0 +1,288 @@
...
+    # HLK may switch the active session user (e.g. DTMLLUAdminUser),
+    # so we detect the current one for the scheduled task principal.
+    def interactive_user
+      user = @tools.run_on_machine(
+        @machine,
+        'Detect interactive session user',
+        '(Get-WmiObject -Class Win32_ComputerSystem).UserName'
+      ).to_s.strip
+      @logger.debug("Interactive session user on #{@machine}: '#{user}'")
+      user.empty? ? 'Administrator' : user.split('\\').last
```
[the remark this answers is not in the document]
kostyanf14: In general, Copilot is correct, but in our case of the HLK env, this is redundant

kostyanf14: Use config['windows_username'] instead

kostyanf14 (changes requested):

lib/models/command_info.rb
YanVugenfirer: @kostyanf14 @elizashurov I agree.

lib/auxiliary/ui_agent/ui_executor.rb
elizashurov: @kostyanf14 config['windows_username'] is always Administrator. After HLK reboots it switches to DTMLLUAdminUser, so we need to detect it dynamically. not?

kostyanf14: `user.empty? ? config['windows_username'] : user.split('\\').last` 
I mean this

lib/auxiliary/ui_agent/ui_executor.rb
elizashurov: I added an -AtLogOn trigger to the scheduled Task and removed the Run key. If the agent stops after a reboot (for example,  when the user switches to from Administrator to DTMLLUAdminUser), run() notices and starts it again automatically.

PR #920, opened 2026-04-03
POC: Notification manager

kostyanf14:
@zjmletang, This is a draft PR, currently, I am not sure how we in RH will use it, so I can create only a general POC. Please review, test, and say what is missing for your idea of usage.

### DIFF
--- a/lib/all.rb
+++ b/lib/all.rb
@@ -56,6 +56,8 @@ module AutoHCK
   autoload_relative :MachineRunError, 'setupmanagers/exceptions'
   autoload_relative :Models, 'models'
   autoload_relative :MultiLogger, 'auxiliary/multi_logger'
+  autoload_relative :NotificationManager, 'notificationmanagers/notification_manager'
+  autoload_relative :NotificationProviderInterface, 'notificationmanagers/notification_provider_interface'
   autoload_relative :NotImplementedError, 'exceptions'
   autoload_relative :OpenJsonError, 'exceptions'
   autoload_relative :PackageManager, 'auxiliary/package_manager'
--- a/lib/all.rb
+++ b/lib/all.rb
@@ -36,6 +36,7 @@ module AutoHCK
   autoload_relative :Downloader, 'auxiliary/downloader'
   autoload_relative :Engine, 'engines/engine'
   autoload_relative :EngineError, 'engines/exceptions'
+  autoload_relative :ExternalScript, 'notificationmanagers/external_script/external_script'
   autoload_relative :ExtraSoftwareBrokenConfig, 'auxiliary/extra_software/exceptions'
   autoload_relative :ExtraSoftwareManager, 'auxiliary/extra_software/manager'
   autoload_relative :ExtraSoftwareMissingConfig, 'auxiliary/extra_software/exceptions'
--- a/lib/engines/config_manager/config_manager.rb
+++ b/lib/engines/config_manager/config_manager.rb
@@ -12,6 +12,7 @@ module AutoHCK
       @project = project
       @logger = project.logger
       @project.append_multilog("#{self.class.tag}.log")
+      @project.notification_manager.post_engine_init(self)
     end
 
     def self.tag(*)
@@ -23,11 +24,14 @@ module AutoHCK
     end
 
     def run
+      @project.notification_manager.pre_engine_run(self)
       ResourceScope.open do |scope|
         @project.logger.info('Stating result uploader token initialization')
         @result_uploader = ResultUploader.new(scope, @project)
         @result_uploader.ask_token
       end
+    ensure
+      @project.notification_manager.post_engine_run(self)
     end
 
     def drivers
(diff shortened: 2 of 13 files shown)


zjmletang:
@kostyanf14 Thanks, this is relevant to my CI use case. I’ll review it and do some validation on my side, then get back to you with feedback. Probably not in the next couple of days, but soon.

zjmletang (commented):
Hi @kostyanf14, thanks for putting this together — the plugin layout and env-var dispatch are clean and easy to extend. I'd like to adopt this in our CI; sharing our two use cases and a bit of feedback around **failure classification** that I hope is useful for the final design.

### Our two use cases

1. **On test completion → trigger downstream automation** (picking up results for driver signing and other post-processing).
2. **On failure → automatically re-run the whole project** (not individual tests). Total wall-time is not critical for us, but minimizing manual intervention is. In practice a large portion of "failures" pass on a plain retry, so a CI-level auto-retry removes a lot of toil.

The granularity we need is **project-level** re-run (much simpler to wire into CI).

### Three failure classes & current hook coverage

From the CI-retry angle, AutoHCK failures really split into three classes:

| # | Class | Typical trigger | Coverage in this PR |
| --- | --- | --- | --- |
| 1 | **Runtime error** (engine / environment layer) | install timeout, studio/client VM failure, engine/tools raising, etc. | ❌ no dedicated event |
| 2 | **Test-level failure** (HLK test ran, status=Failed) | HLK returns Failed result | ✅ already covered by `pre_tests_handle_finished_test_result(s)` |
| 3 | **Hang-on** | test stuck in queued/running, HLK never returns | ❌ AutoHCK detects internally but doesn't surface it |

Class 2 is already actionable with the hooks in this PR (scripts can just read `tests_stats.failed > 0`) — no issue there. The gaps are class 1 and class 3.

### Gap 1 (nice to have) — Class 1: runtime errors have no dedicated event

When the engine aborts **before** reaching the test-handling loop, `pre_tests_handle_finished_test_result(s)` is never reached. `post_project_run` does fire via `ensure`, but:

- the exception is neither passed as an argument nor surfaced in the env;
- `run_terminated` (exposed via `bound_project_args`) only flips when the tracked PR is closed/merged in `github_handling` — it doesn't reflect engine run status.

So scripts can't distinguish "normal run completion" from "engine crashed mid-way". On our side a CI wrapper can work around this by watching the process exit code, so this is only **nice to have**.

**Suggestion (non-blocking)**: if convenient, consider adding a dedicated failure event:

```ruby
sig { abstract.params(project: Project, exception: Exception).void }
def on_project_run_failed(project, exception); end
```

Wired from `Project#run`:

```ruby
def run
  @notification_manager.pre_project_run(self)
  @engine.run
rescue => e
  @notification_manager.on_project_run_failed(self, e)
  raise
ensure
  @notification_manager.post_project_run(self)
end
```

On the `ExternalScript` side, the exception can be exposed via `AUTOHCK_NOTIFICATION_EXCEPTION_CLASS / _MESSAGE / _BACKTRACE`. The change is purely additive — no existing signature is touched.

### Gap 2 (important) — Class 3: "hang-on" is detected but not surfaced

AutoHCK already detects suspected hangs in `Tests#check_test_queued_time` and `Tests#check_test_duration_time` — it logs a warning and sets `ex_status = 'Hangs on at queued state?' / 'Hangs on at running state?'`, but:

- it never raises or aborts, so `ensure`/`rescue`-based notifications never fire;
- **none** of the existing hooks gets called because of a hang;
- external automation has no way to know the run is stuck. We routinely see runs sit in this state for hours until someone kills them.

This matters a lot for CI: you can't auto-retry something that never returns.

**Suggestion**: emit a dedicated notification event when a hang is detected, so operators/CI can decide how to act.

```ruby
sig { abstract.params(test: Models::HLK::Test, state: String, elapsed: Integer).void }
def on_test_hang_detected(test, state, elapsed); end
```

Invoked where `set_test_status(..., 'Hangs on at ...')` is called today; no existing behavior is changed.

**A small caveat**: honestly, I'm not yet sure myself whether the existing hang detection (`QUEUE_TEST_TIMEOUT = 15min`, `RUNNING_TEST_TIMEOUT = 2 × duration + 15min`) covers all the "test getting stuck" cases we care about — for example, situations where HLK keeps reporting Running while the test is actually spinning inside some internal loop may not be caught by the current thresholds. I'll confirm once we start running it in anger.

That doesn't change the ask here though: **exposing what's already detected is the first step**; if additional hang types or new detection logic show up later (progress-stall based, log silence, etc.), they can be layered onto this hook incrementally — either by adding new trigger points or by splitting out another hook. No need to solve everything in one go.

### Summary

- **Class 1 (runtime errors)**: suggest `on_project_run_failed` — **nice to have**, we can work around this with a CI wrapper.
- **Class 2 (test-level failures)**: already covered by `pre_tests_handle_finished_test_result(s)` ✅.
- **Class 3 (hang-on)**: suggest `on_test_hang_detected` —

PR #951, opened 2026-05-11
VIRTWINKVM-2167: add errata support for known HLK test failures

elizashurov:
Allow tests_config entries to specify an errata reason string. When a test fails and matches an errata config, it is treated as "Passed (with errata)" instead of a failure.

- Add Flush Test errata entry for viostor and vioscsi on HLK2022

### DIFF
--- a/lib/engines/hcktest/drivers/vioscsi.json
+++ b/lib/engines/hcktest/drivers/vioscsi.json
@@ -18,6 +18,13 @@
   "tests_config": [
     {
       "tests": [
+        "Flush Test"
+      ],
+      "kits": ["HLK2022"],
+      "errata": "Flush Test - known HLK failure on Win2022"
+    },
+    {
+      "tests": [
         ".*"
       ],
       "pre_test_commands": [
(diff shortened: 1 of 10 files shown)


elizashurov:
Screenshot From 2026-05-10 22-42-52

YanVugenfirer:
Couple of points (not must to fix)
1. Please split to separate commits the functionality and json files next time
2. Maybe little bit different shade of green for tests that passed with errata?

For discussion:
The tests might fail for reasons that are not covered by errata. Let's think how we can verify such cases. Can be done with AI agent comparing elaborated failure description for errata with what is seen on the logs.

kostyanf14:
>The tests might fail for reasons that are not covered by errata. Let's think how we can verify such cases. Can be done with AI agent comparing elaborated failure description for errata with what is seen on the logs.

Manual errata is a magic; we don't know what exactly MS checks.

kostyanf14:
about color
image
This is implemented in Bootstrap, so we need to define a separate color

YanVugenfirer:
Info is good enough color.

elizashurov:
Changed to info color, now looks like: 
image
@YanVugenfirer @kostyanf14

kostyanf14 (approved):

kostyanf14 (approved):

PR #982, opened 2026-06-02
RHELMISC-2846: [SVVP] add sequence_test_names to enforce test execution order

elizashurov:
Some tests require a strict execution order. Add sequence_test_names to svvp_config to move matching tests to the end of the list in the declared order, applied after all other filtering.

### DIFF
--- a/lib/engines/hcktest/hcktest.rb
+++ b/lib/engines/hcktest/hcktest.rb
@@ -131,6 +131,7 @@ module AutoHCK
           'name' => @clients.values[0].name,
           'type' => @svvp_info.type,
           'select_test_names' => @svvp_info.select_test_names,
+          'sequence_test_names' => @svvp_info.sequence_test_names,
           'reject_test_names' => @svvp_info.reject_test_names
         }
       else
@@ -139,6 +140,7 @@ module AutoHCK
           'name' => driver.name,
           'type' => driver.type,
           'select_test_names' => driver.select_test_names,
+          'sequence_test_names' => [],
           'reject_test_names' => driver.reject_test_names
         }
       end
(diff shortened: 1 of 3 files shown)


lib/engines/hcktest/hcktest.rb
```diff
@@ -131,6 +131,7 @@ def target
           'name' => @clients.values[0].name,
           'type' => @svvp_info.type,
           'select_test_names' => @svvp_info.select_test_names,
+          'sequence_test_names' => @svvp_info.sequence_test_names,
```
kostyanf14, line 134: add `'sequence_test_names' => []` to the else block, and make `const :sequence_test_names` not nillable with the default []
then you can drop & from `return unless sequence_names&.any?` or even drop line at all.

lib/engines/hcktest/playlist.rb
```diff
@@ -51,6 +51,8 @@ def list_tests(log)
       custom_select_test_names(log)
       custom_reject_test_names(log)
       sort_by_duration
+      reorder_sequence_tests
+      @tests
```
kostyanf14: `@tests` is redundant 
reorder_sequence_tests returns `@tests`, so list_tests will return reorder_sequence_tests will return `@tests`

elizashurov: `@tests` is necessary because `reorder_sequence_tests` has an early `return if matched.empty? `which returns nil

elizashurov: I can change line to `return @tests if matched.empty?` and then remove `@tests`

kostyanf14 (approved):

PR #991, opened 2026-06-03
RHELMISC 34338: Add client role support (DUT/SUT/Support/Master/Stress)

elizashurov:
Add a role field to platform client config. 
WHQL uses dut/support,SVVP uses sut/master(MC)/stress(SC).

Add `ClientRole` and `PlatformClients` to resolve and validate role
combinations. Master/stress clients skip driver install and HLK target
registration but still join the pool.

### DIFF
--- a/lib/engines/hcktest/hcktest.rb
+++ b/lib/engines/hcktest/hcktest.rb
@@ -8,7 +8,7 @@ module AutoHCK
     extend T::Sig
     include Helper
 
-    attr_reader :config, :drivers, :platform, :extensions
+    attr_reader :config, :drivers, :platform, :extensions, :platform_clients
 
     PLATFORMS_JSON_DIR = 'lib/engines/hcktest/platforms'
     CONFIG_JSON = 'lib/engines/hcktest/hcktest.json'
@@ -17,6 +17,16 @@ module AutoHCK
     SVVP_JSON = 'svvp.json'
     ENGINE_MODE = 'test'
     AUTOHCK_RETRIES = 5
+
+    # Maps platform roles to the HLK machine-set role names used in QueueTest.
+    HLK_ROLE_MAP = {
+      Models::ClientRole::Support => 'Support',
+      Models::ClientRole::Master => 'MC',
+      Models::ClientRole::Stress => 'SC'
+    }.freeze
+
+    # Roles that are the primary test target
+    PRIMARY_ROLES = [Models::ClientRole::Dut, Models::ClientRole::Sut].freeze
 
     def initialize(project)
       @project = project
@@ -28,6 +38,12 @@ module AutoHCK
       @extensions = find_extensions
       prepare_extra_sw
       validate_paths unless @driver_path.nil?
+      @platform_clients = build_platform_clients
+    end
+
+    sig { returns(PlatformClients) }
+    def build_platform_clients
+      PlatformClients.new(@project.engine_platform, logger: @logger)
     end
 
     def test_steps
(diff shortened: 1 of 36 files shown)


elizashurov:
depends on [https://github.com/HCK-CI/rtoolsHCK/pull/81](https://github.com/HCK-CI/rtoolsHCK/pull/81)

lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -147,7 +147,7 @@ def close
     MONITOR_BASE_PORT = 10_000
     VNC_BASE_PORT = 5900
     MAX_RUN_ID = 999
-    MAX_CLIENT_ID = 2
+    MAX_CLIENT_ID = 3
```
kostyanf14: Wow, I forgot to commit 10.
Please add a separate commit to set 10 with a comment like this
```
    MAX_CLIENT_ID = 10
    # https://learn.microsoft.com/en-us/windows-hardware/test/hlk/testref/test-server-configuration
    # 0 - Studio
    # 1 - Active Directory (SVVP-only but mandatory)
    # 2 - SUT / DUT
    # 3 - Master client / Support client for DUT
    # 4 - Stress Client 1
    # 5 - Stress Client 2
    # 6 - Stress Client 3
    # 7 - Stress Client 4
    # 8 - Stress Client 5
    # 9 - Stress Client 6
    # 10 - Stress Client 7 (LoadGen supports up to 64 SC machines, but 8 is usually sufficient, QE uses 4, reserve ID for 7)
```

lib/models/client_role.rb
```diff
@@ -0,0 +1,29 @@
...
+      enums do
+        Dut     = new('dut')     # Device Under Test (WHQL)
+        Sut     = new('sut')     # System Under Test (SVVP)
+        Support = new('support') # Support Client for DUT (WHQL)
+        Master  = new('master')  # Master Client (SVVP)
+        Stress  = new('stress')  # Stress Client (SVVP)
+      end
+
+      sig { params(raw: String).returns(ClientRole) }
+      def self.parse(raw)
```
kostyanf14: Why do we need this? The `class ExecutionState < T::Enum` works without it.

Please add `def to_s` to reduce the use of `.serialize` in other places where we really just need a string.

lib/engines/hcktest/tests.rb
```diff
@@ -521,10 +528,10 @@ def download_memory_dump(machine, l_tmp_path)
...
     def download_memory_dumps(l_tmp_path)
       downloaded_client = download_memory_dump(@client.name, "#{l_tmp_path}/#{@client.name}_#{current_timestamp}")
-      unless @support.nil?
-        downloaded_support = download_memory_dump(@support.name,
-                                                  "#{l_tmp_path}/#{@support.name}_#{current_timestamp}")
-      end
+      downloaded_support = if support_client
+                             download_memory_dump(support_client.name,
+                                                  "#{l_tmp_path}/#{support_client.name}_#{current_timestamp}")
+                           end
```
kostyanf14: Maybe loop for all clients, including stress? I have no idea if it can crash, but it will run HLK software, so theoretically it can.

lib/engines/hcktest/platform_clients.rb
```diff
@@ -0,0 +1,117 @@
+# typed: true
```
kostyanf14, line 1: So cool

lib/engines/hcktest/hcktest.rb
```diff
@@ -434,5 +460,29 @@ def run
...
+
+    sig { returns(T::Hash[String, T::Array[HCKClient]]) }
+    def auxiliary_clients
+      @platform_clients.entries
+                       .reject { |e| PRIMARY_ROLES.include?(e['role']) }
+                       .each_with_object({}) do |e, result|
+                         hlk_role = HLK_ROLE_MAP[e['role']]
+                         client = @clients[e['name']]
+                         (result[hlk_role] ||= []) << client if hlk_role && client
+                       end
```
kostyanf14, line 474: Cool idea, but it this method not only returns clients but also generates something 
maybe rename it or split
or Did I miss some understanding?

lib/models/hlk_platform.rb
```diff
@@ -31,6 +31,7 @@ class HLKClient < T::Struct
       extend JsonHelper
 
       const :name, String
+      const :role, T.nilable(String)
```
kostyanf14, line 34: This is mandatory for QemuHCK, so maybe validate presence?

lib/engines/hcktest/hcktest.rb
elizashurov: Yes better to rename

elizashurov:
@kostyanf14 rebased

PR #1029, opened 2026-07-13
QemuHCK: add --aio-native/--aio-threads CLI options

elizashurov:
Adds CLI flags and setup-manager support for aio=native/aio=threads on
virtual disks. Both modes also set cache=none.

### DIFF
--- a/docs/Home.md
+++ b/docs/Home.md
@@ -92,6 +92,8 @@ Usage: auto_hck.rb test [test options]
         --fs-test-image-format <fs_test_image_format>
                                      Filesystem test image format (qcow2/raw). Default is qcow2.
                                      Has effect only when testing storage drivers.
+        --aio-native                  Use aio=native for virtual disks (forces cache=none, cannot combine with --aio-threads)
+        --aio-threads                 Use aio=threads for virtual disks (forces cache=none, cannot combine with --aio-native)
         --extensions <extensions_list>
                                      List of extensions for run test
         --net-test-speed <net_test_speed>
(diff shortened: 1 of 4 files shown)


lib/engines/functest/functest.rb
```diff
@@ -42,7 +42,10 @@ def self.platform(logger, options)
 
       raise InvalidConfigFile, "Platform configuration not found: #{platform_json}" unless File.exist?(platform_json)
 
-      Models::HLKPlatform.from_json_file(platform_json, logger)
+      platform = Models::HLKPlatform.from_json_file(platform_json, logger)
+      platform.clients_options.drive_aio_state = options.test.drive_aio_state
```
kostyanf14: I think about something like `--fs-test-image-format` but for aio. Is there any reason to make drive_aio_state as a state and add it to the client?  What do you think?

lib/setupmanagers/qemuhck/states.json
```diff
@@ -59,6 +59,20 @@
             ]
         }
     },
+    "drive_aio_state": {
+        "native": {
+            "drive_cache_options": [
+                ",cache=none",
```
kostyanf14: Why force cache = none?

elizashurov: This matches the QEMU command from the Kar tests logs. And according to info from the internet, it is also recommended to use cache=none with aio=native

lib/engines/functest/functest.rb
elizashurov: Agreed, I reworked this

kostyanf14 (approved):

PR #1034, opened 2026-07-15
Functest: Wait for client online before guest commands

rachael-george:
**Issue:**
 `FunctestClient.install_drivers` returned early when the drivers list was empty, skipping the WinRM readiness wait. Any subsequent `guest_run` step would then fail because the VM was not yet reachable.
This surfaces in installer functional tests where the driver is configured with `install_method: "no-drv"`. The driver list is effectively empty after filtering, so the wait was skipped entirely.

**Fix:**
Move the `wait_for_client_online` call before the empty-drivers guard so functest runs without `-d` (or with no-drv drivers) still wait for the guest before executing test steps.

### DIFF
--- a/lib/engines/functest/functest_tools.rb
+++ b/lib/engines/functest/functest_tools.rb
@@ -30,6 +30,13 @@ module AutoHCK
       @logger.info("#{machine} is back after reboot")
     end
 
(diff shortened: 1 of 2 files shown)


kostyanf14:
ok to test

kostyanf14 (approved):

PR #1041, opened 2026-07-16
VIRTWINKVM-1882: Read VM names from config

bhbhatt:
1. Hostnames come from platform JSON — not hardcoded "CL" + number, so custom names  work.
2. One source of truth — Ruby + Windows + HLK use the same client names from the platform.
3. Works with any client count — 1, 2, 3+ clients via args.ps1 ($VMNAMES length matches platform).
4. Lowercase / uppercase JSON both work — install upcases for Windows; test uses casecmp? for HLK.
5. Aligns with Windows/HLK — hostnames are case-insensitive; AutoHCK no longer fails on cl1 vs CL1.
6. Safer for future platforms — change names in JSON + reinstall; no PowerShell formula to update.

### DIFF
--- a/lib/engines/hckinstall/setup-scripts/client.ps1
+++ b/lib/engines/hckinstall/setup-scripts/client.ps1
@@ -34,7 +34,10 @@ function Stage-One {
         if (($macSegments[5] -eq "CC") -and ($macSegments[4] -eq "CC")) {
             $clientNumber = [int32]("0x" + $macSegments[3])
             $clientIp = $clientNumber + 1
-            $clientName = "$CLIENTCOMPUTERNAME$clientNumber"
+            $clientName = $VMNAMES[$clientNumber]
+            if (-not $clientName) {
+                Write-Error "No VM name found in VMNAMES for client number $clientNumber"
+            }
 
             Write-Output "Renaming hostname to $clientName"
             Rename-Computer -NewName "$clientName"
@@ -49,7 +52,7 @@ function Stage-One {
         }
     }
 
-    "$STUDIOIP $STUDIOCOMPUTERNAME #STUDIO VM IP" |  Out-File -encoding ASCII -append 'C:\Windows\System32\drivers\etc\hosts'
+    "$STUDIOIP $($VMNAMES[0]) #STUDIO VM IP" |  Out-File -encoding ASCII -append 'C:\Windows\System32\drivers\etc\hosts'
 
     Enable-PowerShellRemoting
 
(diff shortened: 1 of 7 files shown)


lib/engines/hckinstall/setup-scripts/client.ps1
```diff
@@ -34,7 +34,7 @@ function Stage-One {
         if (($macSegments[5] -eq "CC") -and ($macSegments[4] -eq "CC")) {
             $clientNumber = [int32]("0x" + $macSegments[3])
             $clientIp = $clientNumber + 1
-            $clientName = "$CLIENTCOMPUTERNAME$clientNumber"
+            $clientName = $VMNAMES[$clientNumber]
```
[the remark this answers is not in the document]
kostyanf14: will be useful

lib/engines/hckinstall/setup_scripts_helper.rb
```diff
@@ -71,6 +72,9 @@ def create_setup_scripts_config(workspace_hlk_setup_scripts_path, config)
             value = "'#{v}'"
           when Integer
             value = v
+          when Array
+            items = v.map { |item| "'#{item.to_s.gsub("'", "''")}'" }.join(', ')
```
kostyanf14, line 76: Is there any reason why in example `"` is used (`"STUDIO", "CL1", "CL2"`) while there we wrap strings in `'`? 

```
3.3.7 :001 > list = ["Name1", "Name2", "Name3"]
 => ["Name1", "Name2", "Name3"] 
3.3.7 :002 > list.map { |item| "'#{item.to_s.gsub("'", "''")}'" }.join(', ')
 => "'Name1', 'Name2', 'Name3'" 
```

lib/setupmanagers/hckclient.rb
```diff
@@ -25,9 +25,13 @@ def initialize(setup_manager, scope, studio, name, run_opts)
...
     end
 
+    def machine_name_match?(machine_name)
+      machine_name.casecmp?(@name)
+    end
+
     def pool
       @studio.list_pools.each do |pool|
-        return pool['name'] if pool['machines'].any? { |machine| machine['name'].eql?(@name) }
+        return pool['name'] if pool['machines'].any? { |machine| machine_name_match?(machine['name']) }
```
kostyanf14: Just replace `.eql?` with `casecmp?`, no need for extra function

lib/setupmanagers/hckclient.rb
```diff
@@ -152,8 +156,8 @@ def copy_dvl
     end
 
     def machine_info
-      @studio.list_pools.flat_map { |pool| pool['machines'] }
-                        .detect { |machine| machine['name'].eql?(@name) }
+      machines = @studio.list_pools.flat_map { |pool| pool['machines'] }
+      machines.detect { |machine| machine_name_match?(machine['name']) }
```
kostyanf14: same

lib/engines/hckinstall/setup_scripts_helper.rb
bhbhatt: sorry by mistake will change this to single case

kostyanf14:
ok to test

kostyanf14 (approved):

PR #1044, opened 2026-07-17
Extend driver signature verification to cover full package (.sys, .cat, .inf)

jamepark4:
With https://github.com/HCK-CI/AutoHCK/commit/2f277e0e82e72e535f79d102e481048af6784ecc landing tests have the capacity to install signtool to extend signature verification.

Previously only checked the .sys Authenticode signature. Now also verifies the .cat catalog signature and, when signtool is available via @signtool_path@, cross-checks .sys and .inf hashes against the catalog. Degrades gracefully when signtool is not configured.  Test case json would need to provide expected path to the signtool e.g.:
```
  {
      "name": "driver_sign_check",
      "description": "Verify @driver_module@ driver package signature (.sys, .cat, .inf)",
      "test_system_ref": "VIRT-VSOCK-005",
      "timeout": 60,
      "test_steps": [
          {
              "desc": "Verify driver package signature",
              "guest_run_file": "lib/engines/functest/tests/scripts/verify_driver_signed.ps1",
              "expected_output_contains": "PASS:",
              "timeout": 30,
              "variables": {
                  "@signtool_path@": "@sw_path@\\Signtool_amd64_HLK.exe"
              }
          }
      ]
  }
```

### DIFF
--- a/lib/engines/functest/tests/cases/driver_package_check.json
+++ b/lib/engines/functest/tests/cases/driver_package_check.json
@@ -1,0 +1,13 @@
+{
+    "name": "driver_package_check",
+    "description": "Verify Authenticode signatures of the @driver_module@ driver package (.sys, .cat) before installation",
+    "timeout": 60,
+    "test_steps": [
+        {
+            "desc": "Verify Authenticode signatures of @driver_module@ package",
+            "guest_run_file": "lib/engines/functest/tests/scripts/verify_driver_package.ps1",
+            "expected_output_contains": "PASS:",
+            "timeout": 45
+        }
+    ]
+}
--- a/lib/engines/functest/tests/cases/driver_signtool_check.json
+++ b/lib/engines/functest/tests/cases/driver_signtool_check.json
@@ -1,0 +1,19 @@
+{
+    "name": "driver_signtool_check",
+    "description": "Verify @driver_module@ catalog integrity using signtool — requires extra_software signtool",
+    "timeout": 120,
+    "test_steps": [
+        {
+            "desc": "Locate signtool.exe",
+            "guest_run": "$st = Get-Item (Join-Path $env:SystemDrive '\\signtool\\signtool.exe') -ErrorAction SilentlyContinue; if (-not $st) { $st = Get-ChildItem (Join-Path ${env:ProgramFiles(x86)} 'Windows Kits') -Recurse -Filter signtool.exe -ErrorAction SilentlyContinue | Where-Object { $_.Directory.Name -eq 'x64' } | Select-Object -First 1 }; if ($st) { Write-Output $st.FullName } else { Write-Output 'not-found' }",
+            "capture_output": "signtool_path",
+            "timeout": 30
+        },
+        {
+            "desc": "Verify @driver_module@ catalog integrity via signtool",
+            "guest_run_file": "lib/engines/functest/tests/scripts/verify_driver_signtool.ps1",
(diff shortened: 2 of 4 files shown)


lib/engines/functest/tests/scripts/verify_driver_signed.ps1
```diff
@@ -1,26 +1,57 @@
...
-if (-not $drv) { throw '@driver_module@.inf not found in the DriverStore' }
+if (-not $drv) { throw "${module}.inf not found in the DriverStore" }
 
 $drvDir = Split-Path $drv.OriginalFileName -Parent
-$sys = Get-ChildItem $drvDir -Filter '@driver_module@.sys' | Select-Object -First 1
 
-if (-not $sys) { throw '@driver_module@.sys not found in ' + $drvDir }
+$sys = Get-ChildItem $drvDir -Filter "${module}.sys" | Select-Object -First 1
+$cat = Get-ChildItem $drvDir -Filter '*.cat' | Select-Object -First 1
+$inf = Get-ChildItem $drvDir -Filter "${module}.inf" | Select-Object -First 1
```
kostyanf14: After installation, Windows will move driver.inf to `oem<N>.inf`. This check can work only with the driver binary folder, but not with installed drivers.

@YanVugenfirer Maybe create 2 tests:
1. check installer driver (sys only)
2. check provided binaries that were used for installation (sys, cat, inf)

lib/engines/functest/tests/scripts/verify_driver_signed.ps1
```diff
@@ -1,26 +1,57 @@
...
+
+$sysSig = Get-AuthenticodeSignature $sys.FullName
+Write-Output "SYS: $($sysSig.Status) - Signer: $($sysSig.SignerCertificate.Subject)"
+if ($sysSig.Status -ne 'Valid') {
+    throw "${module}.sys signature invalid: $($sysSig.Status)"
+}
 
-Write-Output "Signature status: $($sig.Status)"
-Write-Output "Signer: $($sig.SignerCertificate.Subject)"
+if ($signtoolPath -eq '@signtool_path@' -or -not (Test-Path $signtoolPath)) {
```
kostyanf14: $signtoolPath = '@signtool_path@'
then $signtoolPath -eq '@signtool_path@' always true
Test-Path is enough

YanVugenfirer:
@kostyanf14 I think we want two tests:

1. Test the driver package before installation. This is the "static" test. 
2. Test installed package(s) with sigverif - this is the test from the OS perspective after installation. Important because we saw issue with all kind of corner scenarios of dual certificates or signatures for specific OSes.

jamepark4:
> @kostyanf14 I think we want two tests:
> 
>     1. Test the driver package before installation. This is the "static" test.
> 
>     2. Test installed package(s) with sigverif - this is the test from the OS perspective after installation. Important because we saw issue with all kind of corner scenarios of dual certificates or signatures for specific OSes.

Thanks @kostyanf14 and @YanVugenfirer based on your feedback I removed any changes to the post installation signature check and create two new pre-install checks. One that does analysis on .sys, .inf, and .cat and one that would leverage signtool if it is supplied to the test suite. I wanted to create two pre-install checks just to allow for better granularity of control and debuggability if there were any issues surrounding using the signtool.

YanVugenfirer (approved):

kostyanf14 (approved):

PR #1046, opened 2026-07-20
QemuHCK: add --discard-granularity CLI option for virtio-blk-pci

elizashurov:
Add a --discard-granularity flag (plain byte count or K/M/G suffix, e.g. 4096, 4K, 256K, 32M). 
Sets discard=unmap on the drive and discard_granularity on the virtio-blk-pci device.
The value is converted to raw bytes before it reaches the QEMU command line.

### DIFF
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -307,7 +307,6 @@ module AutoHCK
       @states_config.each { |name, state| apply_state name, state }
       apply_drive_aio_state
       validate_drive_aio_state
-
       apply_cpu_options_config
     end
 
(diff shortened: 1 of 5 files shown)


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -386,7 +410,9 @@ def options_replacement_map
         '@device_extra_param@' => @device_extra_param.join,
         '@iommu_device_param@' => device_iommu_device_param.join,
         '@cpu_options@' => @cpu_options.join(','),
-        '@drive_cache_options@' => @drive_cache_options.join
+        '@drive_cache_options@' => @drive_cache_options.join,
+        '@drive_discard_param@' => @drive_discard_param.join,
+        '@blk_discard_granularity_param@' => @blk_discard_granularity_param.join
```
kostyanf14: do we really need `@drive_discard_param` and `@blk_discard_granularity_param`? 
We set it only in one place. Maybe go with logic like for viomem (see `def memory_replacement_map`). Create an extra function that will add `,discard=unmap` based on option_config?

elizashurov: Agreed, refactored.

kostyanf14 (approved):

PR #1053, opened 2026-07-21
Add @test_netdev_id@ placeholder and QMP argument substitution

elizashurov:


lib/setupmanagers/qemuhck/qemu_machine.rb
```diff
@@ -364,7 +364,10 @@ def machine_replacement_map
     end
 
     def replacement_map
-      ReplacementMap.new(machine_replacement_map, { '@client_id@' => @client_id })
+      ReplacementMap.new(@project.project_replacement_map, machine_replacement_map, {
+                           '@client_id@' => @client_id,
+                           '@test_netdev_id@' => @nm.test_device_ifname
```
kostyanf14, line 369: is `ifname` =  `id`? I always think `ifname` is not `id`

lib/auxiliary/command_execution_manager.rb
```diff
@@ -231,6 +233,28 @@ def execute_qmp_wait_event(command_info)
...
+    end
+
+    sig { params(value: T.untyped, replacement: ReplacementMap).returns(T.untyped) }
+    def replace_qmp_value(value, replacement)
+      case value
+      when String then replacement.replace(value)
+      when Hash then replace_qmp_arguments(value, replacement)
+      when Array then value.map { |v| replace_qmp_value(v, replacement) }
+      else value
+      end
```
kostyanf14: Let's extend `ReplacementMap` to support different value types
Just extend the `ReplacementMap::replace` function

elizashurov: done

lib/setupmanagers/qemuhck/qemu_machine.rb
elizashurov: The netdev's id is set to `@net_if_name@ ` (same value used for `ifname`)
See here: https://github.com/HCK-CI/AutoHCK/blob/master/lib/setupmanagers/qemuhck/devices/virtio-net-pci.json#L6C36-L6C47

lib/auxiliary/replacement_map.rb
```diff
@@ -41,7 +51,7 @@ def [](key)
     def create_cmd(str)
       result = str.to_s
       @hash.each do |k, v|
-        result = result.gsub(k) { v.shellescape }
+        result = result.gsub(k) { v.to_s.shellescape }
```
kostyanf14, line 54: any reason for to_s there?

kostyanf14 (approved):

lib/auxiliary/replacement_map.rb
elizashurov: @kostyanf14 Because  `replace` now keeps non-strings (e.g integers) as is, we need` .to_s` before .shellescape to avoid errors.

PR #1057, opened 2026-07-22
Add incremental dev id for virtio-blk-pci.json

jamepark4:
Building on the work from [1], updated storage_manager with a dev_id that increments on `device_command_info` call. Similar to [1], update the replacement_map function in `qemu_machine` to also include the formatted device id added to storage_manager. Lastly included new id value in `virtio-blk-pci.json` definition. Tested this downstream against a testcase that also analyzes for configuration properties from the virtio-blk device.

[1] [https://github.com/HCK-CI/AutoHCK/pull/1053](https://github.com/HCK-CI/AutoHCK/pull/1053)


kostyanf14:
ok to test

kostyanf14 (approved):

PR #1066, opened 2026-07-28
QMP: support waiting for multiple alternative events

elizashurov:
qmp_wait_event's event field now also accepts an array of names, returning as soon as any one of them is seen (e.g. pvpanic may emit either GUEST_PANICKED or GUEST_CRASHLOADED).

### DIFF
--- a/lib/auxiliary/command_execution_manager.rb
+++ b/lib/auxiliary/command_execution_manager.rb
@@ -208,10 +208,11 @@ module AutoHCK
       outputs
     end
 
-    sig { params(event: String, machine_name: String, timeout: Integer).returns(T.untyped) }
-    def run_qmp_wait_event(event, machine_name, timeout)
-      @logger.info("Waiting for QMP event '#{event}' on #{machine_name} (timeout: #{timeout}s)")
-      response = @project.setup_manager.wait_for_hypervisor_client_event(machine_name, event, timeout: timeout)
+    sig { params(events: T::Array[String], machine_name: String, timeout: Integer).returns(T.untyped) }
+    def run_qmp_wait_event(events, machine_name, timeout)
+      events_desc = events.join("' or '")
+      @logger.info("Waiting for QMP event '#{events_desc}' on #{machine_name} (timeout: #{timeout}s)")
+      response = @project.setup_manager.wait_for_hypervisor_client_event(machine_name, events, timeout: timeout)
       @logger.debug("QMP event received: #{response.inspect}")
 
       response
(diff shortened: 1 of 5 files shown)


lib/models/command_info.rb
```diff
@@ -53,7 +53,9 @@ class QmpCommandConfig < T::Struct
     class QmpWaitEventConfig < T::Struct
       extend T::Sig
 
-      const :event, String
+      # A single event name, or a list of names to wait for whichever
+      # occurs first (e.g. a device that may emit either one).
+      const :event, T.any(String, T::Array[String])
```
kostyanf14: `T::Array[String]` only. We can rewrite a few tests that already exist, but this will simplify the code.

lib/setupmanagers/qemuhck/qmp.rb
```diff
@@ -38,25 +38,37 @@ def run_cmd(cmd, arguments = nil)
...
+            return response if accepted.include?(response[name])
 
             @events << response if response.key?('event')
             raise(QMPError, response['error'].to_s) if response.key?('error')
           end
         end
       end
 
-      private
-
```
kostyanf14, line 59: why drop private?

elizashurov: I moved private up above above `find_cached_event`

lib/models/command_info.rb
elizashurov: Ok

lib/auxiliary/command_execution_manager.rb
```diff
@@ -208,9 +208,10 @@ def execute_qmp_command(command_info, replacement)
       outputs
     end
 
-    sig { params(event: String, machine_name: String, timeout: Integer).returns(T.untyped) }
+    sig { params(event: T::Array[String], machine_name: String, timeout: Integer).returns(T.untyped) }
```
kostyanf14: event**S** - just because we expect array

lib/models/command_info.rb
```diff
@@ -53,7 +53,9 @@ class QmpCommandConfig < T::Struct
     class QmpWaitEventConfig < T::Struct
       extend T::Sig
 
-      const :event, String
+      # List of event names to wait for; returns as soon as any one of
+      # them occurs (e.g. a device that may emit either one).
+      const :event, T::Array[String]
```
kostyanf14: event**S** - just because we expect array

kostyanf14 (approved):

PR #1069, opened 2026-07-29
[Setupmanager] Create support for two NUMA node guest deployment

jamepark4:
Part of testing `viomem` within tp-qemu were a set of tests that explicitly hotplugged virtio-mem-pci to a guest's second NUMA node.  Besides testing that grow/shrink operations are working it also monitors that the expected NUMA node under test reflects those changes accordingly and NUMA node 0 is not impacted.  HCK-CI currently deploys all guests with a single NUMA node so that testing approach is not possible.  This commit adds support to deploy a two NUMA node guest when requested.  If a driver json definition includes `numa_state: true` then the corresponding guest will deploy with the updated NUMA topology. Implemented the following changes:

File changes:

- `setupmanagers/qemuhck/devices/numa-topology.json`: Created a new device defining default NUMA topology and cmd line.
- `engines/hcktest/drivers/viomem.json`: Updated to with `numa_state: true` and `pcie_spare_root_ports: 1` to allocate an empty PCIe root port for virtio-mem hotplug.
- `models/driver.rb`: Added corresponding properties `numa_state` and `pcie_spare_root_ports`.
- `setupmanagers/qemuhck/devices/virtio-mem-pci.json`: To account for grow/shrink increased maxmem from 2 to 8 for `pluggable_memory_gb`.  Using the variable `pluggable_memory` resulted in the boot device consuming all memory defined by `pluggable_memory_gb` blocking any grow tests. Created a new variable `viomem_boot_backend_size` to maintain the original 2GB size for boot disk and leaving the remaining 6GB for grow operations. Lastly added `viomem_numa_node` to allocate the boot device to NUMA node 0.
- `setupmanagers/qemuhck/qemu_machine.json`: Define default numa state to false
- `setupmanagers/qemuhck/qemu_machine.rb`: If none default values from `numa-topology.json` are not used, dynamically come up with the correct cpu and memory allocation based on total cpu count and memory for the guest. The allocations are purely for a two NUMA node setup. Also update `base_cmd` to use memory slots in the event that NUMA is active.
- `setupmanagers/qemuhck/qemuhck.rb`: Update `OPT_NAMES` with new `numa_state` and `pcie_spare_root_ports` properties.
-  `setupmanagers/qemuhck/states.json`: Update `define_variables` based on `numa_state`.  If true device list now includes numa-toplogy and updates memory_slots to 20, if false define memory_slots to nothing.

### DIFF
--- a/lib/engines/hcktest/drivers/viomem.json
+++ b/lib/engines/hcktest/drivers/viomem.json
@@ -4,5 +4,6 @@
   "inf": "viomem.inf",
   "install_method": "PNP",
   "type": 0,
-  "support": false
+  "support": false,
+  "numa_state": true
 }
--- a/lib/setupmanagers/qemuhck/devices/numa-topology.json
+++ b/lib/setupmanagers/qemuhck/devices/numa-topology.json
@@ -1,0 +1,9 @@
+{
+  "name": "numa-topology",
+  "command_line": [
+    "-object memory-backend-ram,id=numa_mem0,size=@numa_node0_memory@",
+    "-object memory-backend-ram,id=numa_mem1,size=@numa_node1_memory@",
+    "-numa node,memdev=numa_mem0,cpus=0-@numa_node0_max_cpu@",
+    "-numa node,memdev=numa_mem1,cpus=@numa_node1_min_cpu@-@numa_node1_max_cpu@"
+  ]
+}
--- a/lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
+++ b/lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
@@ -1,9 +1,13 @@
 {
   "name": "virtio-mem-pci",
-  "pluggable_memory_gb": 2,
+  "pluggable_memory_gb": 8,
   "need_pci_bus": true,
   "command_line": [
-    "-object memory-backend-ram,id=mem_backend,size=@pluggable_memory@",
-    "-device virtio-mem-pci@device_extra_param@@iommu_device_param@,id=mem_@run_id@_@client_id@,memdev=mem_backend,requested-size=1G,bus=@bus_name@.0"
-  ]
+    "-object memory-backend-ram,id=mem_viomem_@run_id@_@client_id@,size=@viomem_boot_backend_size@",
+    "-device virtio-mem-pci@device_extra_param@@iommu_device_param@,id=mem_@run_id@_@client_id@,memdev=mem_viomem_@run_id@_@client_id@,node=@viomem_numa_node@,requested-size=1G,bus=@bus_name@.0"
+  ],
+  "define_variables": {
+    "@viomem_numa_node@": "0",
+    "@viomem_boot_backend_size@": "2G"
+  }
 }
--- a/lib/models/driver.rb
+++ b/lib/models/driver.rb
@@ -25,6 +25,8 @@ module AutoHCK
       const :s3_state, T.nilable(T::Boolean)
       const :s4_state, T.nilable(T::Boolean)
       const :enlightenments_state, T.nilable(T::Boolean)
+      const :numa_state, T.nilable(T::Boolean)
+      const :pcie_spare_root_ports, T.nilable(Integer)
 
       const :post_start_commands, T::Array[CommandInfo], default: []
       const :extra_software, T::Array[String], default: []
--- a/lib/setupmanagers/qemuhck/qemu_machine.json
+++ b/lib/setupmanagers/qemuhck/qemu_machine.json
@@ -32,6 +32,7 @@
     "vhost_state": true,
     "viommu_state": false,
     "drive_unsafe_cache_state": false,
-    "tpm_state": false
+    "tpm_state": false,
+    "numa_state": false
   }
 }
--- a/lib/setupmanagers/qemuhck/qemu_machine.rb
+++ b/lib/setupmanagers/qemuhck/qemu_machine.rb
@@ -436,6 +436,27 @@ module AutoHCK
       { '@fs_daemon_cache_mode@' => mode }
     end
 
+    def numa_replacement_map
+      return {} unless option_config('numa_state')
+
+      cpu_count = option_config('cpu_count').to_i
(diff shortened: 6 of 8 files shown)


lib/engines/hcktest/drivers/viomem.json
```diff
@@ -4,5 +4,7 @@
   "inf": "viomem.inf",
   "install_method": "PNP",
   "type": 0,
-  "support": false
+  "support": false,
+  "numa_state": true,
+  "pcie_spare_root_ports": 1
```
kostyanf14: What reason for `pcie_spare_root_ports = 1` by default?

lib/setupmanagers/qemuhck/devices/numa-topology.json
```diff
@@ -0,0 +1,16 @@
...
+    "-object memory-backend-ram,id=numa_mem1,size=@numa_node1_memory@",
+    "-numa node,memdev=numa_mem0,cpus=0-@numa_node0_max_cpu@",
+    "-numa node,memdev=numa_mem1,cpus=@numa_node1_min_cpu@-@numa_node1_max_cpu@"
+  ],
+  "define_variables": {
+    "@numa_node0_memory@": "2G",
+    "@numa_node1_memory@": "2G",
+    "@numa_node0_max_cpu@": "3",
+    "@numa_node1_min_cpu@": "4",
+    "@numa_node1_max_cpu@": "7"
```
kostyanf14: I am not sure that we can hardcode the CPU part because it depends on the CPU count in VM. Am I wrong?

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
```diff
@@ -1,9 +1,13 @@
 {
   "name": "virtio-mem-pci",
-  "pluggable_memory_gb": 2,
+  "pluggable_memory_gb": 8,
```
kostyanf14, line 3: Why? Just explain.

kostyanf14 (commented):
Please also check AI comments. Maybe usefull

lib/setupmanagers/qemuhck/devices/numa-topology.json
jamepark4: No you are correct they will always be overridden by the numa_replacement_map, they are unnecessary.

lib/engines/hcktest/drivers/viomem.json
jamepark4: In my testing the default `pcie.0` with the guest did not support hotplugging and I couldn't device_add a pcie-root device at runtime. By default I wanted to pre-allocate one additional empty pcie-root-port at boot for use with the hotplugging.
Edit: It is a supported cli command that can be overwritten though so if the preferred approach is to define it via the cli I can remove it here.

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
jamepark4: Max memory for the guest was being [calculated](https://github.com/HCK-CI/AutoHCK/blob/882e5bb1cf4a7fbe50b83fd7b9588ddddf0863b2/lib/setupmanagers/qemuhck/qemu_machine.rb#L404C11-L404C21) by `memory` and `pluggable_memory_gb`.  The 2GB originally allocated would be for the boot-time virtio-mem backend for viomem driver installation. Given the max memory calculations it prevents any additional memory to be hotplugged.  Increasing the amount just allows for the addition of hotplug device for the test.

lib/setupmanagers/qemuhck/devices/virtio-mem-pci.json
```diff
@@ -1,9 +1,13 @@
...
   "need_pci_bus": true,
   "command_line": [
-    "-object memory-backend-ram,id=mem_backend,size=@pluggable_memory@",
-    "-device virtio-mem-pci@device_extra_param@@iommu_device_param@,id=mem_@run_id@_@client_id@,memdev=mem_backend,requested-size=1G,bus=@bus_name@.0"
-  ]
+    "-object memory-backend-ram,id=mem_viomem_@run_id@_@client_id@,size=@viomem_boot_backend_size@",
+    "-device virtio-mem-pci@device_extra_param@@iommu_device_param@,id=mem_@run_id@_@client_id@,memdev=mem_viomem_@run_id@_@client_id@,node=@viomem_numa_node@,requested-size=1G,bus=@bus_name@.0"
```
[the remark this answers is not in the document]
jamepark4: The configured memory backend for the virtio-mem device for the driver is being replaced from `size=@pluggable_memory@"` to `size=@viomem_boot_backend_size@"`.  The initial `pluggable_memory_gb` is [copied](https://github.com/HCK-CI/AutoHCK/blob/882e5bb1cf4a7fbe50b83fd7b9588ddddf0863b2/lib/setupmanagers/qemuhck/qemu_machine.rb#L403) to `pluggable_memory` but it is no longer being referenced when creating the qemu command for sizing the device.

lib/engines/hcktest/drivers/viomem.json
kostyanf14: I preferred to have it in CLI

kostyanf14:
ok to test

kostyanf14 (approved):

kostyanf14:
@jamepark4 LGTM

kostyanf14:
@jamepark4 

```
lib/setupmanagers/qemuhck/qemu_machine.rb:446

    numa_replacement_map raises generic RuntimeError strings for invalid NUMA configs; the rest of this setupmanager uses QemuHCKError for configuration validation. Also memory_gb should be coerced to an integer the same way as cpu_count so comparisons/division are reliable even if options come in as strings.

      cpu_count = option_config('cpu_count').to_i
      raise "NUMA topology requires at least 2 vCPUs (got #{cpu_count})" if cpu_count < 2

      memory_gb = option_config('memory_gb')
      raise "NUMA topology requires at least 2 GB memory (got #{memory_gb})" if memory_gb < 2
```

Makes sense, but not critical

PR #1090, opened 2026-08-06
[VIRTWINKVM-2679] Fix clean_boot not re-uploading test_binaries to guest

rachael-george:
When a test case uses `clean_boot: true`, `reboot_clean` creates a fresh qcow2 overlay from the base image, which wipes `C:\AutoHCK\test_binaries` from the guest. Tests that reference `@test_binaries_dir@` after a   clean boot fail with "file not found". 

This fix calls `copy_test_binaries` after reconnect in reboot_clean, matching what `prepare_machine` already does during initial setup.

### DIFF
--- a/lib/setupmanagers/functest_client.rb
+++ b/lib/setupmanagers/functest_client.rb
@@ -26,7 +26,7 @@ module AutoHCK
     def reboot_clean
       @logger.info("Rebooting client #{@name} from a clean image")
       @runner = @setup_manager.power_cycle_client(@scope, @name, create_snapshot: true)
-      reconnect
+      prepare_machine(@tools)
     end
 
     # Boots a fresh VM from a previously saved snapshot tag.


kostyanf14:
ok to test

lib/setupmanagers/functest_client.rb
```diff
@@ -27,6 +27,7 @@ def reboot_clean
       @logger.info("Rebooting client #{@name} from a clean image")
       @runner = @setup_manager.power_cycle_client(@scope, @name, create_snapshot: true)
       reconnect
+      copy_test_binaries
```
kostyanf14: What about extra software and drivers?

rachael-george: Good catch! I had only considered the installer test use case when making this fix, where install_drivers is a no-op since we use --test-binaries-path without --driver-path. I'll update the PR to mirror the full prepare_machine sequence: extra software (before driver), drivers, test binaries, and extra software (after driver).

kostyanf14: I think it's better to call `prepare_machine(@tools)`  and drop `reconnect` in this case, instead of duplication

kostyanf14 (approved):

elizashurov (approved):

PR #1112, opened 2026-08-19
qmp: drain stale cached events before waiting for new ones

gayadav:
QEMU emits two DEVICE_DELETED events per device_del (child + parent device). In cycled functest hotplug tests, wait_for consumed only the first event, leaving the second in the cache. On the next cycle, wait_for found the stale cached event and returned immediately without waiting for the current cycle's actual DEVICE_DELETED, allowing device_add to race ahead and fail with "Duplicate device ID".

Change wait_for to drain all matching cached events before blocking on the QMP socket for a fresh event. This ensures each qmp_wait_event step waits for the event from the current operation, not a leftover from a previous cycle.

Verified on Win2016x64_gui (the failing platform) — all 3 hotplug cycles now show real ~1s waits instead of <1ms stale cache hits.


lib/setupmanagers/qemuhck/qmp.rb
```diff
@@ -38,20 +38,17 @@ def run_cmd(cmd, arguments = nil)
...
 
-      # accepted is an array of acceptable values; this returns as soon as
-      # any one of them is seen.
+      # Drain stale matching events from the cache, then block until a
+      # fresh event arrives on the QMP socket.
       def wait_for(name, accepted, timeout = 60)
-        cached = find_cached_event(name, accepted)
-        return cached if cached
-
+        drain_cached_events(name, accepted)
```
elizashurov: Removing these lines can cause test timeouts. It can delete valid events that arrived early and were cached, leaving wait_for waiting for an event that already happened.

elizashurov: Maybe instead of draining the cache, we could tag events with a counter.
We can increment it on every run_cmd call and store the current value with each cached event. Then wait_for would only consider events from the current run and ignore older ones as stale.
That way we don’t need to drain the cache, and we also avoid accidentally dropping an event that just arrived.

elizashurov: @kostyanf14 what do you think?

harshapa-rh: I have added a fix for a similar issue here: https://github.com/HCK-CI/AutoHCK/pull/1099.
Do you want to check if that works, @gayadav ?

kostyanf14: @gayadav @harshapa-rh As you both have a similar issue in the same place. Please work together and propose a fix that covers both use cases.

kostyanf14:
ok to test

kostyanf14:
done in https://github.com/HCK-CI/AutoHCK/pull/1128

PR #1115, opened 2026-08-19
[Engine] Add new install engine for driver installation during guest setup

jamepark4:
A common testcase from the tp-qemu suite is the [unattended_install](https://github.com/autotest/tp-qemu/blob/master/generic/tests/unattended_install.py) which starts up a fresh VM and installs the tested driver(s) during guest setup.  Currently hckinstall supports unattended driver injection during setup but it does not have a phase for functional tests. By comparison functest has phases for both driver installation and functional tests. The issue is the installation phase is done on the live guest rather than during an unattended install.  Created a new engine installtest that shares the installation approach of hckinstall but allows for a follow up phase to conduct functional tests.

What it does at runtime

1. Phase 1 — fresh install with driver. Renders unattended answer files, copies the --driver-path contents into a setup-scripts ISO, and boots each client from a blank image with the setup ISO first (mapped to
   D:) so Windows Setup injects the driver from d:\drivers\. Waits for the unattended install to complete and power down.
2. Phase 2 — boot & verify. Boots the just-installed image(s) as snapshots, connects tools, and runs the selected functest suite (e.g. --category os_install_driver_tests) to confirm the driver is present,
   signed, and the right version. Emits standard HTML/JUnit results.

Invoked as:
`bin/auto_hck installtest -p <platform> --driver-path <dir> -d <driver> --category <suite>`

Changes by area

New engine
- lib/engines/installtest/installtest.rb — InstallTestEngine < FunctestEngine; the two-phase run, install-drive preparation, ISO ordering, and --driver-path as a mandatory hard error (a driver-less install would defeat the engine's purpose).
- lib/engines/installtest/installtest.json — engine config (reuses functest test definitions; install timeout; the autounattend.xml / unattend.xml answer-file list).
- lib/engines/installtest/setup-scripts/client.ps1 — installtest-specific client setup script overlaid on hckinstall's setup-scripts template.

Shared install-during-setup logic
- lib/engines/install_during_setup.rb  — new InstallDuringSetup module holding the answer-file rendering logic (read_iso, product_key_xml, load_fw_disk_config, build_answer_file_path, create_client_answer_files, …), mixed into both HCKInstall and InstallTestEngine.
- lib/engines/hckinstall/hckinstall.rb — refactored to include InstallDuringSetup and drop the now-shared methods. This makes hckinstall's answer-file templates the single source of truth (installtest reuses them, so BIOS and UEFI variants stay in sync) and eliminates ~200 lines of duplication.

Setup-manager support for booting an externally-created image
- qemuhck/storage_manager.rb — new boot_image_override run option: create/boot a snapshot from an arbitrary image path instead of the managed boot image; plus create_boot_image_at(path).
- qemuhck/qemu_machine.rb — boot_image_override default option and create_image_at(path).
- qemuhck/qemuhck.rb — create_client_image_at(name, path) and run_installtest_client(...).
- lib/setupmanagers/installtest_client.rb (new) — InstallTestClient < FunctestClient whose prepare_machine skips driver installation (the driver was already installed during OS setup), doing only the extra-software / test-binary steps.

Registration / CLI wiring
- config.json — "installtest_engine": "installtest".
- lib/all.rb — autoload for InstallDuringSetup, InstallTestEngine, InstallTestClient.
- lib/cli.rb — installtest subcommand (parses the standard test options: -p, -d, --driver-path, --category).
- lib/engines/engine.rb — installtest: InstallTestEngine in the name→class map.

Incidental fix (folded in)
- lib/engines/hcktest/tools.rb — names the block parameter in act_with_tools_on_machine (& → &block). An anonymous block param can't be forwarded from inside a nested block under Ruby 3.3.10; this was a pre-existing bug from the per-machine-locking change that also affects hcktest/functest, surfaced here because Phase 2 is the first to load tools.rb.


kostyanf14:
@jamepark4 Why do we need this? We can already do this with `auto_hck.rb install --drivers <> --driver-path <>`
In this case, drivers will be installed in unattended installation mode. And then just run `auto_hck.rb functest --drivers <> --test-binaries-path <>` 
Two steps, but installation is a separate action in any case.

jamepark4:
> @jamepark4 Why do we need this? We can already do this with `auto_hck.rb install --drivers <> --driver-path <>` In this case, drivers will be installed in unattended installation mode. And then just run `auto_hck.rb functest --drivers <> --test-binaries-path <>` Two steps, but installation is a separate action in any case.

You're correct, there's no real added benefit given the added complexity to make a single command for both the install/functest when it can just as easily be a two phase approach.

jamepark4:
Per https://github.com/HCK-CI/AutoHCK/pull/1115#issuecomment-5352574870, there's no significant benefit of this approach.

PR #1116, opened 2026-08-20
functest: driver_update: fix Win2016 uninstall/reinstall via devcon

elizashurov:
pnputil /delete-driver /uninstall /force and pnputil -i -a are not supported on Windows Server 2016 (build <= 14393). On that build, use pnputil /f /d for store removal followed by devcon remove, and devcon updateni for reinstall, matching tp-qemu behaviour. Newer builds keep the existing pnputil path. Requires the devcon extra_software package.


elizashurov:
@kostyanf14 The AutoHCK run is failing because devcon is missing from extra-software upstream, so I’ll remove this test case from CI.

PR #1121, opened 2026-08-30
S3: Link GitHub statuses directly to results.html report

zjmletang:
Currently, clicking the Details link of the HCK-CI status on a PR opens
the S3 index.html file listing, and one more click is needed to reach the
test report.

This PR makes the link point directly to results.html, so the test
results are easier to reach.

It implements the previously unused html_url hook for the S3 uploader:
the URL is returned once results.html has been uploaded; before that it
returns nil and the existing fallback keeps linking to index.html, so
the link never points to a not-yet-existing object.


kostyanf14:
ok to test

lib/resultuploaders/s3/s3.rb
```diff
@@ -37,7 +38,14 @@ def initialize(project)
       @filenames = []
     end
 
-    def html_url; end
+    # Public URL of the results.html test report, suitable for direct
+    # browser viewing. Returns nil until the report has been uploaded,
+    # so callers fall back to `url` instead of a not-yet-existing object.
+    def html_url
+      return unless @filenames.include?(RESULT_REPORT_FILE_NAME)
```
kostyanf14: Can we use the `AutoHCK::Project::RESULT_REPORT_HTML` constant instead?

zjmletang: done

kostyanf14 (approved):

PR #1126, opened 2026-09-02
HCKTest: Add --package-exclude-unreferenced-files option

elizashurov:
Pass the new exclude_unreferenced_files flag through the tools chain to rtoolsHCK's create_project_package so drivers can be packaged without files not referenced by their INF.


elizashurov:
@kostyanf14 For now, I added this behind the --package-exclude-unreferenced-files flag to avoid breaking the current flow. It depends on [HCK-CI/rtoolsHCK/pull/85](https://github.com/HCK-CI/rtoolsHCK/pull/85).

Once that PR is merged, I’ll add another commit here to bump the rtoolsHCK version.

PR #1127, opened 2026-09-02
QemuHCK: Back default virtio-rng-pci with host /dev/urandom

bhbhatt:
Default -device virtio-rng-pci uses QEMU's builtin RNG, so QEMU never
opens /dev/urandom. rng_host_guest_read needs the guest and host dd to
share that file; without rng-random the guest BAT can still pass but
the concurrent-read check is not real.


kostyanf14:
ok to test

kostyanf14 (approved):

PR #1129, opened 2026-09-03
QemuHCK: add modern virtio-rng device

bhbhatt:
Modern means QEMU must start virtio-rng with disable-legacy=on and disable-modern=off. That is a boot flag.
The default -d viorng device does not set those flags. The test JSON cannot add them after the VM is up.


kostyanf14 (changes requested):
virtio-rng-pci-modern - is not a real device

disable-legacy=on,disable-modern=off should be a CLI option for AutoHCK to add this automatically for proper device

Like: `--discard-granularity`, `--virtio-vectors`, `--aio-threads`.

@YanVugenfirer, what is our decision about modern/legacy testing? 

@bhbhatt if we need this, please rework for all virtio devices that support modern/legacy modes.

bhbhatt:
@YanVugenfirer can you please clarify if we need all three modes or specific modes thanks!

YanVugenfirer:
@bhbhatt As there still no decision on testing of modern and transitional devices, I suggest to migrate existing kar tests

bhbhatt:
https://github.com/HCK-CI/AutoHCK/pull/1148

PR #1134, opened 2026-09-09
CmdRun: join pipe reader threads before close returns

satishnaidu77:
## Summary

`CmdRun#close` waited for the child process with `Process.wait2`, then returned immediately while stdout/stderr were still being read in background threads. Callers that inspect `stddata` right after `close` could see incomplete output.

This showed up in functest host commands (e.g. `ping -c 10`): the command succeeded and printed `PASS:`, but validation ran before the last lines were buffered, causing false failures like `expected to contain 'PASS:'`.

## Fix

- Track stdout/stderr pipe reader threads in `@pipe_threads`
- Join those threads in `close` after `wait2` and before returning

## Test

Add a spec that runs slow stdout and asserts `stddata['stdout']` is complete when `close` returns (no polling). Existing `cmd_run_spec` tests check exit status and logging, but do not cover this race.

## Test plan

- [x] `bundle exec rspec spec/lib/auxiliary/cmd_run_spec.rb` — 8 examples, 0 failures

### DIFF
--- a/lib/auxiliary/cmd_run.rb
+++ b/lib/auxiliary/cmd_run.rb
@@ -11,6 +11,7 @@ module AutoHCK
       @exception = exception
       @logger = logger
       @stddata = { 'stdout' => [], 'stderr' => [], 'both' => [] }
+      @pipe_threads = []
 
       scope.transaction do |transaction|
         ResourceScope.open do |tmp|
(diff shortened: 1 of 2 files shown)


kostyanf14:
ok to test

kostyanf14:
rerun tests

PR #1135, opened 2026-09-10
VIRTWINKVM-2898: Exclude usb-tablet if virtio-mouse/virtio-tablet is attached

harshapa-rh:


kostyanf14 (changes requested):
missing "signoff-by" in commit

kostyanf14:
ok to test

kostyanf14 (approved):

PR #1147, opened 2026-09-21
Add CD-ROM as a device

gayadav:


lib/setupmanagers/qemuhck/devices/cdrom.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "cdrom",
```
kostyanf14, line 2: NO! this is `scsi-cd`, not a cdrom. Why cdrom is `scsi-cd` but not `ide-cd`?

kostyanf14 (changes requested):
You merge adding a new device and some feature "device for test" into one commit. Please split it into separate PRs and explain why you need this feature. 

No problem with the device with proper name

lib/setupmanagers/qemuhck/devices/cdrom.json
gayadav: Sure , I will fix and divide this PR into 2 PRs. And provide information related to requirement of this change

